Issue #114, opened 2026-07-15
CLI config errors should give a CLI-native remediation, not tell an already-CLI user to install @vaultkeeper/cli


mike-north:
## Problem

When the CLI itself surfaces a config error (`ConfigParseError`/`ConfigValidationError`), it prints the *library's* generic multi-audience remediation: "…either install @vaultkeeper/cli and run vaultkeeper config init --force, or repair/replace it programmatically via this library…". A user who is *already running `vaultkeeper`* is told to install the CLI. The wave-2 library→CLI qualification (issue #100) fixed the library-consumer case but flipped the mismatch for the CLI consumer, who is the one that most often hits it (wave-3 A4, confirmed).

## Governing references

- `packages/cli/src/commands/config.ts` (and wherever the CLI catches/prints config errors), `packages/vaultkeeper/src/errors.ts` / `config.ts` (the library's remediation string)

## Acceptance criteria

1. When the CLI surfaces a config parse/validation error, it prints a CLI-native remediation (e.g. "The config at `<path>` is invalid — run `vaultkeeper config init --force` to overwrite it.") — NOT the library's "install @vaultkeeper/cli" text.
2. The library keeps its own @vaultkeeper/cli-qualified/JS-API remediation for library consumers (don't regress issue #100 — the library-only guard test must still pass).
3. The CLI's message still names the actual recovery command (`config init --force`) and the file path.
4. Subprocess regression test: a corrupt config → `config show`/`store` via the CLI prints the CLI-native message and does not tell the user to install the CLI they're already running.

## Non-goals

- Changing recovery behavior (that's `config init --force`, already shipped).

## Proof

The subprocess test in criterion 4; before/after message shown in the PR.

mike-north:
Claiming: dispatching an implementer now to make the CLI print a CLI-native remediation (config init --force + path) for config parse/validation errors, keeping the library's own remediation intact per #100. PR to follow with Refs #114.

mike-north:
Opened https://github.com/mike-north/vaultkeeper/pull/129

mike-north:
Resolved by #129 — CLI config errors now surface CLI-native remediation instead of telling an already-CLI user to install @vaultkeeper/cli.

Issue #100, opened 2026-07-15
Library errors/JSDoc reference a CLI that isn't in the library package; qualify to @vaultkeeper/cli or the JS API


mike-north:
## Problem

Installing only the `vaultkeeper` library (a pure library — no `bin`), a `ConfigParseError` ends with "Run `vaultkeeper config init` to create a valid config", and the `.d.ts` JSDoc references that command in three places (it even cites issue #68). But `npx vaultkeeper` fails ("could not determine executable to run") because the CLI ships separately as `@vaultkeeper/cli`, and the library's packaged README never mentions this. Wave-2 (cohort B) confirmed a library-only user following the error hits a dead end. This is a semantic mismatch: a library's runtime errors must not assume a sibling CLI package is installed.

## Governing references

- `packages/vaultkeeper/src/errors.ts` (`ConfigParseError` and any other error referencing `vaultkeeper config init`), `packages/vaultkeeper/src/config.ts` JSDoc, `packages/vaultkeeper/README.md`
- API-semantics rule (interface honesty).

## Acceptance criteria

1. No library (`vaultkeeper`) runtime error message or public JSDoc instructs the user to run a bare `vaultkeeper config init` as if the CLI were present. Each such reference is either: qualified ("install `@vaultkeeper/cli` and run `vaultkeeper config init`, or construct/repair the config via the JS API"), or replaced with a library-native remediation (e.g. how to pass `config`/`configDir` or write a valid config programmatically).
2. The `vaultkeeper` package README states, once and near the top, that the CLI ships separately as `@vaultkeeper/cli`.
3. A test/grep guard asserts no library error message contains a bare "`vaultkeeper config init`" without the `@vaultkeeper/cli` qualification (or the JS-API alternative).
4. API report regenerated if JSDoc on public symbols changed.

## Non-goals

- Merging the CLI into the library package.

## Proof

The grep/guard in criterion 3; the qualified message shown in the PR.

mike-north:
Claimed for implementation (wave 3) — branching off origin/main (now includes the safe-default work; canonical default is 'file everywhere, native store opt-in').

mike-north:
PR: https://github.com/mike-north/vaultkeeper/pull/110

mike-north:
[...]

Issue #68, opened 2026-07-14
doctor and config show must detect invalid config; parse errors need location and remediation


mike-north:
## Problem

The CLI's diagnostic commands don't diagnose the most common failure mode. With a syntactically invalid `config.json`: `config show` prints the raw garbage verbatim and exits 0; `doctor` reports "System ready." and exits 0 without touching the config. Only `store`/`delete` eventually fail, with `Failed to parse config file at <path>` — no JSON error location, no remediation hint. Related inconsistency: `store` silently succeeds using implicit defaults when *no* config file exists, while `config show` errors with "No config file found." — two commands disagree about whether the system is configured.

Reproduction: `echo '{ bad json' > ~/.config/vaultkeeper/config.json`, then `vaultkeeper config show` (exit 0, raw dump) and `vaultkeeper doctor` (exit 0, "System ready.").

## Governing references

- `packages/cli/src/commands/{config,doctor,store}.ts`; `packages/vaultkeeper/src/config.ts` (loader/validator); `packages/vaultkeeper/src/doctor/`
- README "doctor / preflight" section (doctor's stated purpose is preflight validation).

## Acceptance criteria

1. `doctor` validates that the config file, when present, parses and passes schema validation; an invalid config is reported as a failing check with the parse/validation error and the file path, and doctor exits non-zero.
2. `config show` on invalid JSON exits non-zero with the parse error (including line/position where available) — never a raw dump with exit 0.
3. All config parse errors (including from store/delete/exec) include the file path, the parse location, and a remediation hint naming `vaultkeeper config init`.
4. The no-config behavior is made consistent and documented: pick one story (PM default: commands that need config fall back to platform defaults *and say so* — one line, e.g. "No config file found; using platform defaults (keychain). Run 'vaultkeeper config init' to persist one.") and apply it uniformly across store/delete/exec/config show/doctor.
5. Subprocess regression tests: corrupt-config behavior for doctor/config show/store; no-config messaging for store and config show.

## Non-goals

- Schema changes to config.json.
- General exit-code taxonomy work beyond these commands (separate CLI-polish issue; use exit 1 for these runtime failures consistent with existing convention).

## Proof

Named subprocess tests per criterion; PR shows before/after transcripts for the corrupt-config repro.

mike-north:
[...]

PR #129, opened 2026-07-15
Fix CLI config errors to give a CLI-native remediation, not install @vaultkeeper/cli

mike-north:
## Summary

Refs #114

The CLI was surfacing `ConfigParseError`/`ConfigValidationError`'s raw `.message`, which is the library's generic remediation: "...either install @vaultkeeper/cli and run vaultkeeper config init --force, or repair/replace it programmatically via this library...". A user who is already running `vaultkeeper` (the CLI) was told to install the CLI they already had.

`formatError` (used by every CLI command's catch block: `store`, `delete`, `exec`, `approve`, `rotate-key`, `revoke-key`, `dev-mode`, `doctor`, `config show`/`init`) now special-cases these two error types and builds a CLI-native message from their structured, remediation-free fields (file path, and — for a parse error — its line/column location), rather than reusing `.message`:

> `ConfigValidationError: The config at `<path>` is invalid — run `vaultkeeper config init --force` to overwrite it.`

The library's own message (for JS-API/programmatic consumers, issue #100) is completely unchanged — `packages/vaultkeeper/src/config.ts` and `errors.ts` were not touched.

1. **CLI prints a CLI-native remediation, not "install @vaultkeeper/cli"**
   - `packages/cli/test/unit/output.test.ts` → `formatError > config errors get a CLI-native remediation (issue #114) > rewrites ConfigParseError...` / `...ConfigValidationError...`
   - `packages/cli-tests/test/e2e/config.test.ts` → `should exit non-zero with path, parse location, and remediation hint for corrupt config.json (issue #68 repro)` and `should exit non-zero with path and remediation hint for a structurally invalid config.json` (both now assert `.not.toContain('install @vaultkeeper/cli')`)

2. **Library keeps its own remediation for library consumers (no regression of issue #100)**
   - Unchanged: `packages/vaultkeeper/test/unit/cli-reference-guard.test.ts` (still passing, untouched) and `packages/vaultkeeper/test/unit/config.test.ts`'s existing assertions on `err.message` containing `vaultkeeper config init` — library source files were not modified.

3. **CLI's message still names the recovery command and file path**
   - Same tests as (1); each asserts `.toContain(configPath)` and `.toContain('vaultkeeper config init --force')`.

4. **Subprocess regression test: corrupt config → `config show`/`store` prints CLI-native message**
   - `packages/cli-tests/test/e2e/config.test.ts` (config show, both parse and validation error cases)
   - `packages/cli-tests/test/e2e/store-delete.test.ts` → `store should exit non-zero with a CLI-native remediation (path, parse location, config init --force) for corrupt config.json (issues #68, #114)`

## Before / after

Before:
```
ConfigValidationError: Invalid config at /home/user/.config/vaultkeeper/config.json: Config version must be 1. Fix the file — either install @vaultkeeper/cli and run 'vaultkeeper config init --force' to overwrite it with a valid config, or repair/replace it programmatically via this library (pass an explicit `config` or `configDir`, or write a valid config.json yourself).
```

After:
```
ConfigValidationError: The config at `/home/user/.config/vaultkeeper/config.json` is invalid — run `vaultkeeper config init --force` to overwrite it.
```

## Non-goals / scope notes

- Recovery behavior (`config init --force`) is unchanged — this only changes the message shown before recovery.
- Unit tests in `packages/cli/test/unit/commands/*.test.ts` that used a full `vi.mock('vaultkeeper', () => ({...}))` replacement had to switch to a partial mock (`importOriginal`) so `formatError`'s `instanceof` checks against the real `ConfigParseError`/`ConfigValidationError` classes still work under mocking — this is a mechanical test-infra fix, not a behavior change.
- **Known related gap, deliberately out of scope**: `doctor`'s "config" preflight check (`packages/vaultkeeper/src/doctor/runner.ts`) stores the raw library error message as `PreflightCheck.reason`, which `doctorCommand` prints verbatim in `nextSteps` — so `vaultkeeper doctor` against a corrupt config still shows the library's "install @vaultkeeper/cli" text today. Fixing that well requires exposing structured error info on `ScopedPreflightCheck` (a public API shape change), which is a larger design decision beyond this issue's stated acceptance criteria and governing files. Recommend a follow-up issue.

## Test plan
- [x] `pnpm check` (typecheck + lint + api-report) green
- [x] `pnpm test` green (249/249 in `@vaultkeeper/cli`, full workspace green)
- [x] Prettier run on all changed files

### DIFF
--- a/.changeset/cli-native-config-remediation.md
+++ b/.changeset/cli-native-config-remediation.md
@@ -1,0 +1,5 @@
+---
+'@vaultkeeper/cli': patch
+---
+
+Fixed the CLI printing the library's "install @vaultkeeper/cli" remediation when it hit an invalid config (`ConfigParseError`/`ConfigValidationError`) — a user already running this CLI was told to install a CLI they already had. The CLI now prints its own remediation naming the file path and the actual recovery command: "The config at `<path>` is invalid — run `vaultkeeper config init --force` to overwrite it." The library's own message (used by JS-API consumers) is unchanged.
--- a/packages/cli-tests/test/e2e/config.test.ts
+++ b/packages/cli-tests/test/e2e/config.test.ts
@@ -176,6 +176,15 @@ describe('config command', () => {
   // Repro from issue #68: a syntactically invalid config.json must never be
   // dumped verbatim with exit 0 — it must fail with the parse error, the
   // file path, a parse location, and a remediation hint.
+  //
+  // Issue #114: the remediation hint must be CLI-native ("run `vaultkeeper
+  // config init --force`"), not the library's "install @vaultkeeper/cli"
+  // text — a user running this CLI already has it installed. The library's
+  // own field-level validation reason (e.g. "version must be 1") is no
+  // longer echoed verbatim, since it travels only inside the library's
+  // message alongside that wrong remediation (issue #100); the CLI instead
+  // builds its own message from the error's structured, remediation-free
+  // fields (path, and — for a parse error — its line/column location).
   it('should exit non-zero with path, parse location, and remediation hint for corrupt config.json (issue #68 repro)', async () => {
     env = await createCliTestEnv()
     await fs.writeFile(path.join(env.configDir, 'config.json'), '{ bad json', 'utf8')
@@ -184,7 +193,8 @@ describe('config command', () => {
     expect(result.stdout).toBe('')
     expect(result.stderr).toContain(path.join(env.configDir, 'config.json'))
     expect(result.stderr).toMatch(/line \d+, column \d+/)
-    expect(result.stderr).toContain('vaultkeeper config init')
+    expect(result.stderr).toContain('vaultkeeper config init --force')
+    expect(result.stderr).not.toContain('install @vaultkeeper/cli')
   })
 
   it('should exit non-zero with path and remediation hint for a structurally invalid config.json', async () => {
@@ -197,8 +207,8 @@ describe('config command', () => {
     const result = await env.run(['config', 'show'])
     expect(result.exitCode).not.toBe(0)
     expect(result.stderr).toContain(path.join(env.configDir, 'config.json'))
-    expect(result.stderr).toContain('version must be 1')
-    expect(result.stderr).toContain('vaultkeeper config init')
+    expect(result.stderr).toContain('vaultkeeper config init --force')
+    expect(result.stderr).not.toContain('install @vaultkeeper/cli')
   })
 
   it('should exit 2 for config with no subcommand', async () => {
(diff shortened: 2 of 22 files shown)


mike-north:
Requesting changes on two in-scope items. Everything else here is solid — the layer choice (CLI-side formatting built only from structured, remediation-free fields; library untouched per #100) is right, and the unit + e2e coverage is thorough.

## 1. The config-read failure path still prints the wrong-audience message (blocking)

`loadConfig` throws `FilesystemError` for an unreadable `config.json` (EACCES/EPERM), and its message embeds the same `CONFIG_REMEDIATION_HINT` as the two handled classes (`packages/vaultkeeper/src/config.ts:374`). `formatError` has no case for it, so e.g. `vaultkeeper store` against a root-owned config still prints "…install @vaultkeeper/cli and run 'vaultkeeper config init --force'…" — the exact defect #114 targets, on a third error path.

It's detectable without string matching: the error carries structured `path` and `operation: 'read'`, and every `formatError` call site now passes `configDir`, so `err instanceof FilesystemError && err.operation === 'read' && err.path === path.join(configDir, 'config.json')` identifies it precisely (FileBackend's reads are under the backend storage dir, never `configDir/config.json`, so there's no collision). Note the remediation for this case shouldn't be `config init --force` — that would fail on the same permission error. Something like "check the file's permissions" is the honest hint.

Please add a subprocess e2e case for it alongside the existing corrupt-config ones (chmod 000 the config file, skip on Windows).

## 2. Validation errors lost their diagnostic detail (blocking, ~one line)

Before: `Invalid config at …: Config version must be 1 …`. After: ``The config at `…` is invalid — run `vaultkeeper config init --force` to overwrite it.`` The user is steered to a destructive overwrite without being told what's wrong, and `ConfigValidationError.field` — exactly the kind of structured, remediation-free property `formatConfigError`'s own docstring says it relies on — goes unread.

Including it (e.g. ``The config at `<path>` is invalid (`version`) — run …``) recovers most of the lost diagnostic value for one line. Relatedly, `packages/cli/test/unit/commands/config.test.ts` dropped its only field-specific assertion (`toContain('version must be 1')`) with no replacement — once `field` is in the message, assert it so the diagnostic can't silently regress. (Exposing the full human-readable *reason* as a structured field on the error class is a real API change and correctly out of scope; `field` gets us most of the way.)

## Non-blocking

- The doctor gap you flagged is tracked in #130 — I've added a note there that the wrong text currently prints twice per `doctor` run (check reason + nextSteps).
- The partial `vi.mock('vaultkeeper', importOriginal)` boilerplate is now copy-pasted across 7 command test files — filing a follow-up to extract a shared mock factory.
- Optional while you're in here: `path.join(configDir, 'config.json')` is hand-derived in 5 places across the CLI and library; a tiny `configJsonPath(configDir)` helper in the CLI would at least collapse the three CLI sites (and finding 1 above adds a use).

Issue #136, opened 2026-07-15
TS-version matrix: release.yml never installs the pinned compilers, and CI skips should be failures


mike-north:
## Problem

#128 merged the TS-version compatibility matrix with the fixture-install step in `ci.yml` only. Two gaps remain (flagged in the post-merge review of #128; also the substance of the one legitimate unresolved Copilot thread there):

1. **`release.yml`'s `ci` job also runs `pnpm test`** (after the same build/check sequence) but never installs `packages/vaultkeeper/test/e2e/ts-version-fixtures` — so during release validation all four matrix cells register as vitest skips. The release pipeline does not verify the shipped `.d.ts` against the supported TypeScript range, and the test header's "CI always installs the fixtures first, so every cell actually runs there" is untrue on the release path.
2. **In CI, a missing pinned compiler skips the cell instead of failing.** `it.skipIf` made local runs honest, but CI skips are invisible-by-default: any future workflow refactor that drops or reorders the install step silently turns the whole gate off (item 1 proves this drift mode is real). The cell should skip only outside CI and fail loudly inside it, e.g.:

```ts
const inCI = process.env.CI !== undefined
it.skipIf(!typescriptRoot && !inCI)(`typechecks under ${label}`, async () => {
  if (!typescriptRoot) {
    throw new Error(
      `${label}: pinned compiler not installed — run pnpm --dir packages/vaultkeeper/test/e2e/ts-version-fixtures install --ignore-workspace (CI must run this before pnpm test)`,
    )
  }
  ...
```

## Acceptance criteria

1. `release.yml`'s test-running job installs the fixtures with the same step as `ci.yml` (`pnpm --dir packages/vaultkeeper/test/e2e/ts-version-fixtures install --frozen-lockfile --ignore-workspace`) before `pnpm test`.
2. Matrix cells fail (with the remediation message) when the compiler is missing and `process.env.CI` is set; still skip cleanly locally.
3. Optional hardening while in here: each cell asserts the resolved compiler's real version (from `<typescriptRoot>/package.json`) matches its pin, so cell labels and the READMEs' "5.0.4–7.0.2" claim are verified rather than assumed.

## Non-goals

Changing the pinned version set or the fixtures-install mechanism (verified working as-is — `--ignore-workspace` after the command is fine).

## Proof

Release-workflow test logs show 4/4 matrix cells executed; deleting the install step locally with CI=1 makes the cells fail rather than skip.

Refs #128, #125

[...]

PR #144, opened 2026-07-15
Close the TS-version matrix gate for release validation

mike-north:
## Summary

Refs #136

`release.yml`'s `ci` job runs `pnpm test` but never installed the pinned TypeScript compilers `consumer-typecheck.test.ts`'s matrix needs, so all four matrix cells silently registered as Vitest skips during release validation — the release pipeline never actually verified the shipped `.d.ts` against the supported TypeScript range. Separately, a missing pinned compiler skipped the cell everywhere (including CI), so a future workflow refactor that dropped the install step would silently disable this gate with no red test to catch it.

- `.github/workflows/release.yml`'s `ci` job now installs `packages/vaultkeeper/test/e2e/ts-version-fixtures` with the exact same step `ci.yml` uses, right before `pnpm test`.
- `consumer-typecheck.test.ts`'s skip/fail decision is extracted into `ts-version-matrix-gate.ts` as `shouldSkipMatrixCell(typescriptRoot, isCI)`: a missing compiler only skips **outside** CI. Inside CI (`process.env.CI` set) the cell now runs and throws an actionable `missingCompilerMessage(label)` naming the install command, instead of skipping.
- The test file's header comment (previously claiming "CI always installs the fixtures first, so every cell actually runs there") is corrected to describe the new skip-outside-CI / fail-inside-CI behavior.
- `ts-version-matrix-gate.test.ts` adds unit coverage for the extracted predicate and the remediation message, independent of `tsc` or a real CI environment.

## Acceptance criteria → evidence

1. **`release.yml` installs the fixtures with the same step as `ci.yml`.** See the diff in `.github/workflows/release.yml` — identical command (`pnpm --dir packages/vaultkeeper/test/e2e/ts-version-fixtures install --frozen-lockfile --ignore-workspace`), same comment, placed before the (renamed for accuracy) `Test (including conformance and TS-version matrix)` step.
2. **Matrix cells skip only outside CI; inside CI a missing compiler fails with a remediation message.** Verified locally:
   - Fixtures installed, `pnpm test` (no `CI` env): all 4 cells run and pass.
   - Fixtures removed, no `CI` env: all 4 cells report as Vitest **skips** (unchanged local behavior).
   - Fixtures removed, `CI=1`: all 4 cells **fail**, each with `<label>: pinned compiler not installed — run pnpm --dir packages/vaultkeeper/test/e2e/ts-version-fixtures install --frozen-lockfile --ignore-workspace (CI must run this before pnpm test)`. Output below.
3. **Header comment corrected.** `consumer-typecheck.test.ts`'s module doc now states the skip is CI-conditional and explains why (see diff).
4. **Unit-level verification.** `shouldSkipMatrixCell` and `missingCompilerMessage` are extracted to `ts-version-matrix-gate.ts` and unit-tested in `ts-version-matrix-gate.test.ts` (6 cases: skip/run decision for all `{compiler present/absent} × {CI on/off}` combinations, plus the remediation message content and exact install command).

### `CI=1` with fixtures absent (the regression this closes)

```
$ rm -rf packages/vaultkeeper/test/e2e/ts-version-fixtures/node_modules
$ CI=1 pnpm --filter vaultkeeper exec vitest run test/e2e/consumer-typecheck.test.ts

 ❯ test/e2e/consumer-typecheck.test.ts (4 tests | 4 failed)
   × ... typechecks under TypeScript 5.0.4 (stated floor)
     → TypeScript 5.0.4 (stated floor): pinned compiler not installed — run pnpm --dir packages/vaultkeeper/test/e2e/ts-version-fixtures install --frozen-lockfile --ignore-workspace (CI must run this before pnpm test)
   × ... typechecks under TypeScript 5.9.3 (latest 5.x)
     → TypeScript 5.9.3 (latest 5.x): pinned compiler not installed — ...
   × ... typechecks under TypeScript 6.0.3 (latest 6.x)
     → TypeScript 6.0.3 (latest 6.x): pinned compiler not installed — ...
   × ... typechecks under TypeScript 7.0.2 (latest 7.x)
     → TypeScript 7.0.2 (latest 7.x): pinned compiler not installed — ...

 Test Files  1 failed (1)
      Tests  4 failed (4)
```

### Fixtures installed (normal path)

```
$ pnpm --dir packages/vaultkeeper/test/e2e/ts-version-fixtures install --frozen-lockfile --ignore-workspace
$ pnpm --filter vaultkeeper test

 ✓ test/unit/ts-version-matrix-gate.test.ts (6 tests)
 ✓ test/e2e/consumer-typecheck.test.ts (4 tests)
 Test Files  46 passed (46)
      Tests  734 passed (734)
```

`pnpm check` also passes clean (one pre-existing, unrelated lint warning in `packages/cli/src/config-dir.ts`, not touched by this PR).

## Non-goals

Not changing the pinned TypeScript version set or the fixtures-install mechanism — both verified working as-is. No changeset: this is CI/test-infrastructure only, no published-package behavior changes.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -82,7 +82,16 @@ jobs:
       - name: Make binary executable
         run: chmod +x target/release/vaultkeeper
 
-      - name: Test (including conformance)
+      # Standalone, non-workspace install of the pinned TypeScript compiler
+      # versions used by the consumer-typecheck TS-version matrix (#125).
+      # Kept out of pnpm-workspace.yaml — see that directory's package.json
+      # for why — so it needs its own install step here. Mirrors ci.yml's
+      # step exactly: release validation must run the same matrix as CI,
+      # not silently skip it (#136).
+      - name: Install TS-version matrix compilers
+        run: pnpm --dir packages/vaultkeeper/test/e2e/ts-version-fixtures install --frozen-lockfile --ignore-workspace
+
+      - name: Test (including conformance and TS-version matrix)
         run: pnpm test
 
   release:
--- a/packages/vaultkeeper/test/e2e/consumer-typecheck.test.ts
+++ b/packages/vaultkeeper/test/e2e/consumer-typecheck.test.ts
@@ -39,11 +39,20 @@
  * the conformance runner's `describe.skipIf` pattern) — not a silent early
  * `return`, which Vitest would otherwise report as a false pass — so local
  * runs without that setup step show honest "skipped" cells instead of
- * misleadingly green ones; CI always installs the fixtures first, so every
- * cell actually runs there.
+ * misleadingly green ones.
+ *
+ * That skip is outside-CI only. Both `ci.yml` and `release.yml` install the
+ * fixtures before running `pnpm test`, so inside CI (`process.env.CI` set) a
+ * missing pinned compiler is never treated as a skip — the cell still runs
+ * and fails with a remediation message naming the install command. This
+ * means a future workflow refactor that drops or reorders the install step
+ * is caught by a failing test rather than silently turning this gate off
+ * (#136). See `ts-version-matrix-gate.ts` for the extracted skip/fail
+ * predicate and `test/unit/ts-version-matrix-gate.test.ts` for its coverage.
  *
  * @see https://github.com/mike-north/vaultkeeper/issues/72
  * @see https://github.com/mike-north/vaultkeeper/issues/125
+ * @see https://github.com/mike-north/vaultkeeper/issues/136
  */
 
 import { describe, it, expect, afterEach } from 'vitest'
@@ -53,6 +62,7 @@ import { promisify } from 'node:util'
 import { createRequire } from 'node:module'
 import { resolve, dirname } from 'node:path'
 import { fileURLToPath } from 'node:url'
+import { shouldSkipMatrixCell, missingCompilerMessage } from './ts-version-matrix-gate.js'
 
 const execFileAsync = promisify(execFile)
 const __dirname = dirname(fileURLToPath(import.meta.url))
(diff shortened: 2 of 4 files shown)


mike-north:
This implements both required criteria from #136 faithfully — the release.yml step mirrors ci.yml exactly, the skip/fail decision is a pure, fully-quadrant-tested predicate, and the remediation message carries the actual install command. One fix before merge:

## The workflow-consistency test is tautological

`remediation message names a command that matches the CI install step exactly` asserts `FIXTURES_INSTALL_COMMAND` equals a string literal that is a copy of the constant. It can only fail if someone edits the constant — it does not verify the thing its name claims: that the **workflow files** actually run this command. The drift #136 guards against is precisely a workflow edit, and this test would stay green through it.

Make it real by reading the workflows and asserting both contain the command:

```ts
import { readFileSync } from 'node:fs'

it.each(['.github/workflows/ci.yml', '.github/workflows/release.yml'])(
  '%s runs the fixtures-install command the remediation message names',
  (workflow) => {
    const contents = readFileSync(resolve(repoRoot, workflow), 'utf8')
    expect(contents).toContain(FIXTURES_INSTALL_COMMAND)
  },
)
```

(`repoRoot` via `resolve(__dirname, '../../../..')` from the test file, same pattern as the config.ts source-audit test.) That closes the loop: renaming/dropping the step or changing its flags now fails a unit test naming the exact command to restore, in addition to the in-CI cell failure — and the in-CI failure only triggers *after* a bad refactor ships to a workflow run, whereas this catches it at PR time.

The deferred optional version-assertion with a TODO pointing back at #136 is a reasonable scoping call, and `Refs` (not `Closes`) keeps that tracked. Everything else looks right — I'll merge once this lands.

Issue #200, opened 2026-07-16
CLI config init writes trustTier as a JSON number, but the WASM/Rust-core reader requires a string — WASM SDK can't read a CLI-produced config


mike-north:
## Problem (wave-11 study A2 — CONFIRMED, cross-package interop break on a documented path)

`vaultkeeper config init --backend file` writes `"trustTier": 3` (bare JSON number), matching the README's Configuration example. But the Rust-core config deserializer (used by the WASM SDK) expects trustTier as a STRING-encoded number per the documented serde contract (TrustTier serializes as "1"/"2"/"3"). So `await createVaultKeeper()` reading that same config dir throws `VaultError: Failed to parse config: expected value at line 14 column 18` (pointing at the `3`). A config directory set up via the documented CLI command is unusable by the WASM SDK out of the box — despite the SDK being pitched as matching native-CLI behavior. Editing that one field to `"3"` makes it parse.

## Repro (confirmed, wave-11 tarballs)

1. `npm i vaultkeeper-cli-0.1.9.tgz vaultkeeper-wasm-0.2.1.tgz`
2. `HOME=<iso> node ./node_modules/@vaultkeeper/cli/dist/bin.js config init --backend file`  → writes `"trustTier": 3`
3. `node -e "import('@vaultkeeper/wasm').then(m => m.createVaultKeeper())"`  → `VaultError: Failed to parse config: expected value at line 14 column 18`

## Acceptance criteria

Pick ONE canonical wire form and make CLI writer + Rust-core reader + README example all agree, AND (preferred) make the reader lenient so existing configs of both forms load:

1. Make the Rust-core config deserializer accept a trustTier that is EITHER a JSON number OR a string-encoded number (serde number-or-string), so both a CLI-produced `3` and a `"3"` parse. (Most robust / backward-compatible.)
2. Align the CLI `config init` output and the README Configuration example with the canonical documented form (the serde contract is string `"3"`) — OR, if you make the reader lenient and keep the CLI emitting a number, update the README example + the TrustTier serde docs so the whole ecosystem is internally consistent about what is written vs accepted.
3. Add a CONFORMANCE case that round-trips a real `config init` output through the WASM SDK's `createVaultKeeper()` (and the TS library's `init`) so this can never regress silently. Both the CLI (number today) and WASM must agree.
4. Confirm the TS library reader and the CLI reader also handle whatever canonical form is chosen (no new break there).

[...]

PR #205, opened 2026-07-16
Fix trustTier config interop: CLI writes a number, WASM/core reader now accepts it

mike-north:
Refs #200

## Problem

`vaultkeeper config init --backend file` writes `"trustTier": 3` (a bare JSON number), matching the README Configuration example and the TS library. But the Rust-core config deserializer used by the WASM SDK required a **string-encoded** number (`"3"`), so `await createVaultKeeper()` reading a config dir produced by the documented CLI flow threw:

```
VaultError: Failed to parse config: expected value at line 14 column 18
```

A config directory set up via the documented `config init` command was unusable by the WASM SDK out of the box.

### Repro (before)

```
NUMBER form: ERR -> Failed to parse config: expected value at line 1 column 134
STRING form: OK
```

### After

```
NUMBER form: OK
STRING form: OK
```

## Fix

Canonical wire form chosen: **bare JSON number** (`3`) — it already matched the CLI, TS library, and README; only the Rust-core reader (and native-CLI writer) disagreed.

- **Reader is lenient**: `defaults.trustTier` now deserializes from either a bare number (`3`) or a string-encoded number (`"3"`), so both old and new configs load. Scoped to the config field via a local serde adapter, so the `tid` claim in JWE tokens keeps its string wire form (`"3"`) for `jose` compatibility.
- **Writer emits a number**: the field now serializes as a bare number, so the native Rust CLI's `config init` output matches the TS CLI, the TS library, and the README — the two CLIs are once again format-identical.
- The TS library/CLI reader and writer already used the numeric form (TS reader was already lenient); no change needed there.

## Tests

- **Rust core** (`unit_tests.rs`): numeric form parses, string form still parses (backward compat), out-of-range tier rejected, and `default_config()` serializes `trustTier` as a bare number. The existing `trust_tier_serializes_as_string_number` test (JWE `tid`) still passes, proving the change is scoped.
- **WASM SDK** (`sdk.test.ts`): loads a config whose `trustTier` is a bare number.
- **Cross-package e2e** (`config-interop.test.ts`, new): round-trips a **real** `config init` output through the TS library `loadConfig()` **and** the WASM SDK `createVaultKeeper()` — so a CLI-writer / any-reader divergence can never regress silently. Asserts the on-disk form is a number, never a string.

## Rebuilt artifact

The fix lives in `vaultkeeper-core`, so the committed WASM binary was rebuilt with `wasm-pack build --target nodejs` — only `vaultkeeper_wasm_bg.wasm` changed (JS bindings and `.d.ts` are byte-identical, no interface drift).

## Verification

- `cargo test` (all crates), `cargo fmt --all --check`, `cargo clippy --all-targets` — clean
- `pnpm --filter @vaultkeeper/wasm test` — 72 pass
- full `pnpm check` (typecheck + lint + format + api-report) and `pnpm check:api-docs` — clean
- full `pnpm test` — 202 pass

Changeset: `@vaultkeeper/wasm` patch.


mike-north:
Reviewed and approving. The one risk that mattered — the JWE wire format — is verified isolated: the `trust_tier_config` adapter is applied via `#[serde(with)]` to `VaultDefaults::trust_tier` only (spot-checked myself), `TrustTier` keeps its `"1"/"2"/"3"` string renames for the `tid` claim, and the full Rust suite including token round-trips passes. The committed wasm binary was rebuilt from this branch's source and came out byte-identical, both writer paths (Rust CLI init + dev-mode) now emit the bare-number form, the cross-package UAT exercises the *real* TS CLI writer through both readers, and CLAUDE.md's interop note was corrected to document the two-wire-form split instead of the now-false blanket claim.

Two non-blocking nits for a future pass, verified but not gating: (1) `-1`/`1.5`/`null` fail with serde's generic untagged-enum message rather than the friendly out-of-range one (nothing panics or silently succeeds — just uneven messaging, and those three shapes have no tests); (2) the string branch's `s.trim()` whitespace leniency is undocumented and untested, so it reads as incidental.

Merging when CI is green.

Issue #179, opened 2026-07-15
README follow-ups from #175: false 'repository README mirrors this' claims and missing pack-contents sentinel check


mike-north:
## Problem

Two review findings on #175 merged unaddressed (the blocking doctor-scoping fix did land):

1. **False "mirrors" claims.** `packages/vaultkeeper/README.md` (~355, ~380) and `packages/cli/README.md` (~145) say the [repository README] "mirrors this list/content for online reference" — but the root `README.md` contains none of #175's additions: no "Multiple secrets in one request" section, no Ed25519/Ed448 verify() caveat, no full `VaultConfig`/`BackendConfig` field table, and its error-types table is missing at least 5 of the 25 classes (`InvalidKeyMaterialError`, `DecryptionError`, `FetchError`, `ConfigValidationError`, `ConfigParseError`). A reader clicking through for the promised online copy finds a stale subset.
2. **No pack-contents content check.** #172's AC explicitly asked that the shipped tarball's README completeness be verified by a pack-contents test. `packaging.test.ts` still only asserts `README.md` exists in the tarball — nothing prevents a future edit from silently stripping the guaranteed sections.

## Acceptance criteria

1. Either sync the root README so the mirrors claim is true (add the missing sections/rows), or reword the three claims to not promise parity (e.g. "the repository README covers related narrative online").
2. `packages/cli-tests/test/packaging/packaging.test.ts` asserts the packed `vaultkeeper` README contains sentinel strings for the guaranteed content — e.g. "Multiple secrets in one request", `InvalidKeyMaterialError`, `gracePeriodDays`, and "Doctor / preflight checks" — so stripping any of them fails CI.
3. Umbrella-convention check while in there: root README references the main package; package READMEs reference themselves.

Refs #175, #172, #171

mike-north:
Claiming: README tree quiet post-#159; dispatching an implementer for the mirrors-claim fix + pack-contents sentinel test. PR to follow with Refs #179.

mike-north:
PR opened: https://github.com/mike-north/vaultkeeper/pull/207

mike-north:
Shipped in #207 via option (b): the three "mirrors" claims are reworded to name the package README as the authoritative reference, and the pack-contents sentinel test landed stronger than specified — it packs the real tarball and extracts the shipped README (not a dry-run file list), asserting all four guaranteed-section sentinels. Verified locally (24/24 packaging tests). This can close. Leaving the close to the PM lane.

[...]

PR #207, opened 2026-07-16
Fix false README parity claims and add pack-contents sentinel check

mike-north:
## Summary

- Rewords the three "the repository README mirrors this list/content" claims in `packages/vaultkeeper/README.md` (error hierarchy, full documentation section) and `packages/cli/README.md` (full documentation section). Re-verified against the current `main` (post-#159, post-#196): the root README is missing an entire section ("Multiple secrets in one request"), the full `VaultConfig`/`BackendConfig` field table, and 4 of 29 documented error classes (`DecryptionError`, `FetchError`, `ConfigValidationError`, `ConfigParseError`). Since the gap spans whole sections rather than a couple of missing rows, this reworks the claims to point readers to the package's own README as the authoritative reference instead of duplicating that content into a second hand-maintained file — a parity promise between two files is a standing drift liability, which is exactly what this issue is proof of.
- Confirmed no other "mirrors"/parity claims exist elsewhere (grepped both READMEs plus `@vaultkeeper/wasm`'s — its one "see the repository README" pointer references real, non-duplicated content and isn't a parity claim, so it's untouched).
- Confirmed the umbrella-convention check already holds: the root README's "Which package should I use?" section references `vaultkeeper`/`@vaultkeeper/wasm`/`@vaultkeeper/cli`, and each package README is titled with its own name and links back to `vaultkeeper` only where relevant. No violations found, no changes needed.
- Adds a `packages/vaultkeeper` README-content packaging test: packs the package for real (not `--dry-run`), extracts `README.md` from the resulting tarball, and asserts it contains sentinel strings for guaranteed content — the "Multiple secrets in one request" heading, `InvalidKeyMaterialError`, `gracePeriodDays`, and the "Doctor / preflight checks" heading. Verified this fails when a sentinel is stripped (tested locally by removing one, confirming the new test failed, then reverting).

No changeset — docs and tests only, no shipped package behavior changed.

## Test plan

- [x] `pnpm check` (typecheck + lint + api-report) — green
- [x] `pnpm test` — 175 passed, 25 skipped (Rust-conformance tests that skip without a built binary, unrelated)
- [x] `pnpm exec prettier --check` on changed files — green
- [x] Manually verified the new packaging test fails when a guaranteed README section is stripped, then passes again once restored

Refs #179

### DIFF
--- a/packages/cli-tests/test/packaging/packaging.test.ts
+++ b/packages/cli-tests/test/packaging/packaging.test.ts
@@ -15,7 +15,8 @@
 import { describe, it, expect } from 'vitest'
 import { execFile } from 'node:child_process'
 import { promisify } from 'node:util'
-import { readFile } from 'node:fs/promises'
+import { mkdtemp, readFile, rm } from 'node:fs/promises'
+import * as os from 'node:os'
 import * as path from 'node:path'
 import { fileURLToPath } from 'node:url'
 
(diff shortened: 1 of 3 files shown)


mike-north:
Reviewed and approving — this closes #179 properly, and the sentinel test is implemented *better* than the issue suggested: instead of grepping a dry-run file list, it packs for real and extracts `package/README.md` from the actual tarball, so it proves the shipped bytes contain the guaranteed sections (ran it: 24/24 packaging tests pass, sentinel test green at ~0.5s). All three "mirrors" claims are reworded to the honest "covers related narrative online" framing with the package README named as authoritative — which also future-proofs against the root README drifting again. Type-guarded `npm pack --json` parsing follows the repo's no-`as`-casts conventions.

Note for the close: this implements #179's option (b) — softening the claims — rather than syncing the root README to make them true. That's the right call (the package README being self-authoritative is the stabler invariant), and #179 can close with this.

Merging when CI is green.

Issue #208, opened 2026-07-16
Flaky real-clock race in jwe/claims.test.ts ('accepts future expiry one second from now') — use fake timers


mike-north:
Surfaced during wave-12 CI (#203). packages/vaultkeeper/test/unit/jwe/claims.test.ts 'accepts future expiry one second from now' uses the real clock (Date.now()), so it intermittently fails when the 1-second window elapses between setup and assertion (observed: 'Token expired at 1784161356, now 1784161356'). Violates the fixed-dates-in-tests convention. Fix: use vi.useFakeTimers()/vi.setSystemTime() to control the clock deterministically instead of real Date.now(). Audit sibling time-dependent tests in the file for the same pattern. Low priority but causes CI flakiness (real failures require a re-run, masking genuine ones).

mike-north:
Claiming: dispatching an implementer to convert the real-clock expiry test to fake timers per the repo's fixed-dates test convention. PR to follow with Refs #208.

mike-north:
Fixed in https://github.com/mike-north/vaultkeeper/pull/218

mike-north:
Merged in PR #218. The real-clock race in jwe/claims.test.ts ('accepts future expiry one second from now') is gone: the test now runs under vi.useFakeTimers()/vi.setSystemTime with a fixed date constant (repo convention), restoring real timers in afterEach; determinism verified by five consecutive full-file runs. The rest of the file was audited — the remaining time-based tests use large past-dated margins or fail before the clock check, so they cannot flip and were deliberately left on real time. Test-only, no changeset.

PR #218, opened 2026-07-16
Fix flaky real-clock race in claims.test.ts future-expiry test

mike-north:
## Summary

- `jwe/claims.test.ts`'s `accepts future expiry one second from now` read `Date.now()` once to build a claim expiring 1 second out, then let `validateClaims()` read `Date.now()` again independently. If the real clock crossed a second boundary between those two reads, the claim appeared expired, flaking CI (per issue's observed failure: `Token expired at 1784161356, now 1784161356`).
- Converted the test to use `vi.useFakeTimers()` / `vi.setSystemTime()` (fixed date) with `vi.useRealTimers()` in `afterEach`, scoped to a nested `describe` block, matching the fake-timer convention already used elsewhere in the suite (e.g. `test/unit/keys/storage.test.ts`).
- Audited the rest of the file for the same flake shape (a small real-clock margin that shrinks between claim construction and validation). No other test has it:
  - `throws TokenExpiredError when exp is in the past` / `exp equals now` / `TokenExpiredError has canRefresh=false` all use a large (3600s) **past**-dated margin — as real time advances the claim only gets *more* expired, so these can't flip results.
  - `throws VaultError when iat is after exp` fails on the `iat > exp` check before the clock-based expiry check is ever reached, so it's clock-independent.
  - Nothing left behind requiring conversion.

No changeset included — this is a test-only change with no user-facing effect.

## Test plan

- [x] `pnpm --filter vaultkeeper exec vitest run test/unit/jwe/claims.test.ts` run 5x in a row — 30/30 tests pass every time, confirming determinism
- [x] `pnpm check` — typecheck, lint, api-report all green
- [x] `pnpm test` — full suite green (181 passed, 25 skipped as expected — conformance tests skip without the Rust binary)
- [x] `pnpm exec prettier --check` on the changed file — clean

Refs #208

### DIFF
--- a/packages/vaultkeeper/test/unit/jwe/claims.test.ts
+++ b/packages/vaultkeeper/test/unit/jwe/claims.test.ts
@@ -1,4 +1,4 @@
-import { describe, it, expect, beforeEach } from 'vitest'
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest'
 import { validateClaims, blockToken, isBlocked, clearBlocklist } from '../../../src/jwe/claims.js'
 import type { VaultClaims } from '../../../src/types.js'
 import {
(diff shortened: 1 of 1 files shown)


mike-north:
Reviewed inline — approving. This is precisely the right fix shape per the repo's deterministic-time testing conventions: vi.useFakeTimers() + vi.setSystemTime() with a fixed instant so validateClaims()'s independent Date.now() read can't cross a second boundary against the test's, real timers restored in afterEach, and the regression comment explains the #208 race rather than papering over it with a wider tolerance. Merging when CI is green.

Issue #217, opened 2026-07-16
Add a README-code-fence CI check that runs/type-checks shipped examples against the built packages (prevents example drift)


mike-north:
## Motivation

Across the DX study waves, the single most recurring class of finding has been **README/docs example drift** — a code or CLI example in a shipped README that doesn't match the actual built package (the wave-13 sign/verify `printf` vs `<<<` bug; repeated WASM quick-start confusion; missing `--config-dir`; earlier PEM/DER and signature wording). These are exactly the issues a fresh blind user hits first, and they keep recurring because nothing mechanically verifies the examples.

## Proposal

Add a CI check that EXTRACTS the fenced code/command examples from the shipped READMEs and RUNS or TYPE-CHECKS them against the built packages:
- **CLI fences** (```bash / ```sh in @vaultkeeper/cli + root README): execute the documented command sequences in an isolated HOME + file backend and assert they succeed (exit 0 where the doc implies success), so a sign→verify or store→exec walkthrough that doesn't actually work fails CI.
- **TS/JS fences** (```ts / ```js in vaultkeeper + @vaultkeeper/wasm READMEs): compile them under the repo's strict NodeNext tsconfig against the built .d.ts (and optionally run the runnable ones), so a snippet with a wrong signature (e.g. a 2-arg setup()) fails CI.

## Acceptance criteria
1. A test/CI job that parses shipped READMEs for fenced examples and validates them against the built packages, wired into `pnpm check` or a dedicated `check:readme-examples` target run in CI.
2. It catches the wave-13 sign/verify example bug (fails before the fix, passes after) — use that as the proving case.
3. Examples that are intentionally illustrative-not-runnable can be opted out with an explicit marker (e.g. a fence info-string or a preceding HTML comment), documented so authors know how.
4. Reasonable scope: it need not run every fence, but must cover the getting-started / quick-start / sign-verify / access-pattern examples in each shipped README.

This is a durable prevention for the dominant finding class. Source: wave-13 confirm study (quick-wins) + cross-wave pattern.

mike-north:
Resolved by #225 — a README example-fence CI check (readme-example-harness.ts + readme-examples.test.ts) extracts shell fences and runs them against the built CLI (isolated HOME + file backend) and type-checks TS/JS fences against the built .d.ts; illustrative fences opt out via ''. It immediately caught that #214's fix was half-applied and flagged fragment fences needing markers — exactly the example-drift class it prevents.

Issue #214, opened 2026-07-16
CLI README sign/verify example fails as written (sign uses printf, verify uses <<< — trailing-newline mismatch)


mike-north:
## Problem (wave-13 study A2 + A3 — CONFIRMED by two verifiers, 100% deterministic on bash/zsh)

The @vaultkeeper/cli README's own copy-pasteable "Signing keys, sign, and verify" walkthrough fails as written. It feeds the payload to `sign` via `printf '%s' "$CHALLENGE"` (no trailing newline) but to `verify` via the here-string `<<<"$CHALLENGE"`, which bash AND zsh terminate with an appended `\n`. Sign and verify therefore operate on byte-different payloads (e.g. 15 vs 16 bytes), so the detached EdDSA signature legitimately fails with "Signature did not verify." (exit 3). The crypto is correct — this is an internally inconsistent doc example on the tool's flagship offline-verifiable feature.

## Repro (confirmed)

Isolated HOME + file backend:
1. `vaultkeeper key create --name approval-signing-key --type ed25519`
2. `vaultkeeper key export --name approval-signing-key > approval.pub`
3. `CHALLENGE="hello-challenge"; printf '%s' "$CHALLENGE" | vaultkeeper sign --name approval-signing-key > sig`
4. `vaultkeeper verify --public-key approval.pub --signature sig <<<"$CHALLENGE"` → "Signature did not verify.", exit 3

Control: feeding verify the same way as sign — `printf '%s' "$CHALLENGE" | vaultkeeper verify ...` → "Signature valid.", exit 0.

## Acceptance criteria

1. Fix the README sign/verify example so `sign` and `verify` receive byte-identical stdin — feed BOTH via `printf '%s' "$CHALLENGE"` (piped), OR add an explicit inline note that `<<<` appends a trailing newline and the payloads must match byte-for-byte. Prefer the consistent-construct fix so a verbatim copy-paste succeeds.
2. Verify the corrected example runs end-to-end (sign → verify → "Signature valid.", exit 0) as written in bash AND zsh.
3. Check for the same `printf` vs `<<<` inconsistency anywhere else the README demonstrates stdin-fed commands (store, exec).

Source: wave-13 confirm study, subjects A2 + A3. See also the companion README-fence CI check issue.

mike-north:
[...]

PR #232, opened 2026-07-16
Fix README sign/verify example and add README example-fence CI check

mike-north:
Refs #214, Refs #217

## Summary

Fixes the `@vaultkeeper/cli` README `sign`/`verify` walkthrough (#214) and adds a CI check that runs/type-checks the shipped README examples against the built packages so this class of drift can't recur (#217).

## #214 — sign/verify README example fails as written

The walkthrough fed the payload to `sign` with `printf '%s'` (no trailing newline) but to `verify` with a here-string (`<<<`), which bash **and** zsh terminate with an appended `\n`. `sign` and `verify` therefore covered byte-different payloads, so the detached EdDSA signature legitimately failed with `Signature did not verify.` (exit 3). Both commands now read byte-identical stdin via `printf '%s' "$CHALLENGE" | vaultkeeper …`, with an inline note explaining the newline trap. The same fix is applied to the root README's sign/verify block. Verified end-to-end in **bash and zsh** (`Signature valid.`, exit 0).

### `@vaultkeeper/cli` README: Signing keys, sign, and verify

```diff
+CHALLENGE="hello-challenge"
 vaultkeeper key create --name approval-signing-key --type ed25519   # unknown --type exits 2
 vaultkeeper key export --name approval-signing-key > approval.pub   # SPKI PEM public key
 printf '%s' "$CHALLENGE" | vaultkeeper sign --name approval-signing-key > sig
-vaultkeeper verify --public-key approval.pub --signature sig <<<"$CHALLENGE"   # exit 0 = valid
+printf '%s' "$CHALLENGE" | vaultkeeper verify --public-key approval.pub --signature sig   # exit 0 = valid
```

## #217 — README example-fence CI check

A new `check:readme-examples` target validates the fenced examples in the shipped READMEs against the **built** packages:

- **Shell fences** (`sh`/`bash`) in the CLI + root README are executed verbatim through real `bash -euo pipefail` in an isolated `HOME`/config dir + `file` backend, with a `vaultkeeper` shim on `PATH`. Running the actual shell is what reproduces the #214 byte-semantics — the sign/verify walkthrough **fails this check before the fix (exit 3) and passes after** (proven by temporarily reverting the README fix).
- **TS/JS fences** (`ts`/`js`) in the `vaultkeeper` + `@vaultkeeper/wasm` READMEs are type-checked under the repo's strict NodeNext config against each package's built `.d.ts`, so a snippet with a wrong signature fails CI.

Intentionally-illustrative fences (fragments that reuse a `vault` from an earlier fence, placeholder tool paths, etc.) opt out with a documented HTML-comment marker: `<!-- readme-example: skip - <reason> -->`. Package-manager install lines (`pnpm add`, `cargo install`, …) are auto-skipped since they need network.

Wired as an uncached nx target (`check:readme-examples`), added to the aggregate `pnpm check`, and run as its own CI step. Harness + tests live in `packages/cli-tests` (`test/e2e/readme-example-harness.ts`, `readme-examples.test.ts`), which already does isolated-HOME subprocess CLI testing.

## Verification

- `pnpm check:readme-examples` passes on the fixed READMEs; fails on a reverted #214 fix (fence exits 3).
- `pnpm --filter @vaultkeeper/cli-tests test` — all pass.
- `pnpm check` — green (typecheck, lint, format, api-report, readme-examples).

## Changeset

`@vaultkeeper/cli` patch (the README sign/verify fix). The CI check is dev-only (private `@vaultkeeper/cli-tests`), so no changeset there.


mike-north:
Superseded by #225 (CLI README sign/verify fix + fence check) and #226 (root README coverage + changeset), both already merged to main. This branch is a duplicate of that landed work. Closing and deleting the branch.

Issue #235, opened 2026-07-18
Rust core must default to the 'file' backend on every platform (never silently target the keychain)


mike-north:
Refs #234 (consolidation Phase 1) · Refs #98

## Problem

`packages/vaultkeeper/src/config.ts` deliberately hard-codes `'file'` as the zero-config default backend on **every** platform — the #98 fix: "a missing config must never silently target the real keychain." But `crates/vaultkeeper-core/src/config/mod.rs::default_config()` does the opposite: `keychain` on macOS, `dpapi` on Windows, `file` only elsewhere. When the TS package swaps its internals to the wasm core (epic #234 Phase 3), this divergence would silently reintroduce the #98 regression — new users' secrets written to the OS keychain by default. This must be resolved and pinned now, before any engine swap.

## Acceptance criteria

1. `default_config()` in `crates/vaultkeeper-core` returns the `file` backend as the zero-config fallback on macOS, Windows, and Linux. The platform-native selection logic is removed from the default path (explicit config may still select keychain/dpapi).
2. A Rust regression test asserts the fallback backend type is `file` on all three platforms (parameterize over the platform enum rather than cfg-gating so all arms are exercised on any host). Test name/comment references #98.
3. A conformance case is added to `crates/vaultkeeper-conformance` asserting `config show` (or the closest equivalent surface) reports the `file` backend in a fresh config-less environment, so both runners enforce it.
4. The wasm-exposed config path (`@vaultkeeper/wasm`) reflects the same default; existing TS-side tests for `defaultBackendType()` remain untouched and green.

## Non-goals

- No change to explicit-config behavior or the CLI `config init` wizard.
- No broader config-validation porting (typed `UnknownBackendTypeError`/`ConfigParseError` parity is separate work).

## Proof

`cargo test` green including the new regression test; conformance runners pass; `pnpm check` and full TS tests green.

mike-north:
Claimed for implementation (consolidation Wave A, epic #234). Branch: issue-235-file-default.

mike-north:
Opened https://github.com/mike-north/vaultkeeper/pull/250

mike-north:
Delivered by PR #250 (merged as 4fdbe31). Acceptance criteria met:

1. `default_config()` returns the `file` backend on every platform; platform-native selection removed from the zero-config path (explicit config still selects keychain/dpapi).
[...]

Issue #234, opened 2026-07-18
Epic: single-core consolidation — one Rust core behind TS (wasm) and the native CLI


mike-north:
## Decision (maintainer-approved 2026-07-17)

vaultkeeper consolidates on a single semantic core: `crates/vaultkeeper-core`.

- TypeScript consumers reach the core through WASM (`crates/vaultkeeper-wasm` → `packages/vaultkeeper-wasm`).
- Native consumers reach the core through the Rust CLI.
- The parallel pure-TS business logic in `packages/vaultkeeper/src` is **retired as a semantic authority**. The `vaultkeeper` npm package keeps its public API, surviving as a host layer (Node I/O adapters, typed-error reconstruction, ergonomics).
- The TS CLI (`packages/cli`) is retired in favor of npm-distributing the native binary via per-platform packages.

**Standing freeze rule, effective immediately: new semantic behavior lands in `vaultkeeper-core` first.** TS-first feature work that widens the parity gap is not accepted while this migration is in flight.

## Why

Parallel implementations produced recurring security-relevant divergence, all caught by review rather than tooling (wasm errno erasure, `verify_trust` never wired into wasm, TOFU manifest clobber fixed twice, trust-type divergence, #75 CLI drift). Parity cost is per-feature forever; bridge cost is mostly one-time. Full decision record: `docs/specs/005-single-core-consolidation.md` (maintainer-side spec set).

## Trust boundary (governs all child issues)

The host is the trusted embedder; the adversary is a third party holding a JWE token. Everything defending against that adversary — key material, JWE decrypt/validate, expiry, usage limits, executable-trust binding, signing-vs-secret discrimination, presence *refusal* decisions — lives in the core. Backend capability claims are host-attested (documented limit, not a defended one).

## Phase map

- **Phase 0 — contracts/foundations** (gate): HostPlatform extensions, HostSecretBackend contract, wasm build guards, pinning tests.
- **Phase 1 — core parity** (parallel): signing stack, key persistence, presence/capabilities, CapabilityToken handle table, delegated-access policy, config default resolution.
- **Phase 2 — backend inversion**: JsSecretBackend dispatch; keychain/dpapi/secret-tool/yubikey become core-over-exec; 1Password stays a JS host-callback backend.
- **Phase 3 — engine swap**: `packages/vaultkeeper` internals delegate to the core, subsystem by subsystem, gated on the **unchanged** existing TS test suites passing with zero test edits.
[...]

Issue #98, opened 2026-07-15
Make the safe file backend the default (or unreachable-keychain guard): zero-config must not silently write to the real OS keychain


mike-north:
## Problem

The zero-config default still silently targets the user's **real OS keychain** on macOS (DPAPI on Windows). Wave-2 confirmed `VaultKeeper.init({skipDoctor:true}).activeBackendType === 'keychain'` with no config file, and the documented copy-paste quick start / `config init` land there. The opt-out now works (`config init --backend file`, explicit file config) and the CLI prints a *post-hoc* hint, but a first-time user following the documented first example can unknowingly write a secret into their real login keychain before they know backends exist. The `doctor`/`store` inline remediation hint ("Run `vaultkeeper config init` to persist one") also omits `--backend`, so following it verbatim persists the keychain backend.

This is the residual of wave-1 blocker #7 — the dangerous default itself, reachable by copy-paste.

## Governing references

- `packages/vaultkeeper/src/config.ts` (default config generation), `packages/cli/src/commands/{config,doctor,store}.ts`, root `README.md` quick start + the `[!WARNING]` block, per-package READMEs
- API-semantics rule: the safe, portable, CI-friendly behavior should be the one a user reaches by default; a destructive-to-the-real-keychain action should require explicit opt-in.
- Wave-2 study report.

## Acceptance criteria

Choose ONE of the following two designs and implement it fully (state which in the PR):

**Design A (preferred — safe default):** a bare `VaultKeeper.init()` / `config init` with no backend specified defaults to the **file** backend. The platform keychain becomes explicit opt-in (`config init --backend keychain`, or explicit config). Update all docs, the `[!WARNING]`, and the platform-defaults table to match.

**Design B (guarded default):** keep the platform default, but make it impossible to reach the real keychain by copy-pasting the documented first example — e.g. the quick start and `config init` require an explicit `--backend` choice (no silent default), and `store` refuses to write to a keychain backend that was never explicitly chosen.

Either way:
1. The shortest documented getting-started path cannot silently write a secret to the real OS credential store.
2. The `doctor`/`store` "no config" remediation hint shows `config init --backend file` explicitly (not a bare `config init`).
3. README quick start (root + packaged) shows the safe path first; the platform-default/keychain behavior is documented as the explicit opt-in.
4. `activeBackendType` introspection and existing backend-selection tests updated; a test asserts the documented first example does not resolve to keychain.

## Non-goals

- Removing keychain support (it stays, as opt-in).
[...]

PR #250, opened 2026-07-18
Fix Rust core to default to the file backend on every platform (#98/#235)

mike-north:
Refs #235

## Problem

`crates/vaultkeeper-core/src/config/mod.rs::default_config()` defaulted to `keychain` on macOS and `dpapi` on Windows, diverging from `packages/vaultkeeper/src/config.ts`'s #98 fix, which deliberately hard-codes `file` as the zero-config default on every platform. A missing config must never silently target the real OS keychain — this divergence would silently reintroduce the #98 regression once the wasm core swap (#234 Phase 3) lands.

## Fix

`default_config()` in `crates/vaultkeeper-core` now always resolves to the `file` backend, regardless of platform. Explicit configuration that selects `keychain`/`dpapi`/`secret-tool` is unaffected — only the zero-config fallback changed. A new `default_backend_type_for_platform(Platform) -> &'static str` helper (always returning `file`) lets tests iterate every `Platform` arm explicitly instead of relying on `cfg!(target_os = ...)`, which only exercises one arm per host.

1. **`default_config()` returns `file` on macOS/Windows/Linux; explicit config still allows keychain/dpapi** → `crates/vaultkeeper-core/src/config/mod.rs` (implementation); explicit-config behavior untouched (`config_validation` tests in `crates/vaultkeeper-core/tests/unit_tests.rs` around parsing an explicit `"keychain"` backend still pass).
2. **Rust regression test parameterized over the `Platform` enum, referencing #98** → `crates/vaultkeeper-core/tests/unit_tests.rs::config_validation::default_config_is_file_backend_on_every_platform` (iterates `Platform::Darwin/Linux/Windows`). Also strengthened `crates/vaultkeeper-cli/tests/cli_integration.rs::config_init_uses_file_backend_on_every_platform`, which previously had the same platform-dependent assumption baked into `#[cfg(target_os = ...)]` blocks.
3. **Conformance case asserting `config show` (or closest equivalent) reports `file` in a fresh config-less environment, enforced by both runners** → extended the `"config init"` case in `crates/vaultkeeper-conformance/src/lib.rs` with a new `expected_config_file` assertion (a `JsonContains` check against the written `config.json`), wired into both `crates/vaultkeeper-conformance/tests/run_conformance.rs` (Rust runner) and `packages/cli-tests/test/conformance/run-conformance.test.ts` (JS runner reading `cases.json`, regenerated via `cargo run -p vaultkeeper-conformance --bin export-conformance`).
4. **wasm-exposed config path reflects the same default; existing TS `defaultBackendType()` tests untouched** → `@vaultkeeper/wasm`'s `VaultKeeper` shares the same `vaultkeeper_core::config::default_config()`/`load_config()` code path, so the fix propagates automatically. Added `packages/vaultkeeper-wasm/src/test/sdk.test.ts::resolves the file backend as the zero-config default with no config.json present (#98)` as an explicit regression test. `packages/vaultkeeper/src/config.ts` (`defaultBackendType()`) was not touched.

## Other changes

- Rebuilt and committed the wasm artifact (`packages/vaultkeeper-wasm/wasm/`) since the Rust core changed.
- Added a patch changeset for `@vaultkeeper/wasm` (its observable zero-config default backend changed).
- `Cargo.lock` picked up an unrelated pre-existing version drift (0.2.1 → 0.3.0) when rebuilding — the lockfile was stale relative to the workspace version already committed elsewhere.

## Proof

- `cargo test --workspace`: all green (67+68+29+8+1+3 tests across crates).
- `cargo clippy --workspace --all-targets`: clean.
- `cargo fmt --all -- --check`: clean.
- `pnpm check`: green (pre-existing lint warnings only, no errors).
- `pnpm test`: 241 passed, 19 skipped (unrelated, pre-existing skips), across all TS packages including the `@vaultkeeper/wasm` `node:test` suite (73 passed).

## Non-goals (per issue)

- No change to explicit-config behavior or the CLI `config init` wizard.
- No broader config-validation porting (typed error parity) — separate work.


mike-north:
Rebased onto main (wasm-guards baseline) and rebuilt the committed artifact with the pinned toolchain; the only conflict was the binary. Full Rust workspace + wasm SDK suites green on the rebased tree. Branch was force-pushed — authoring session please pull before further work.

Issue #278, opened 2026-07-21
Implement profile resolution planner and materialization engine in vaultkeeper-core


mike-north:
Refs #276

Implement the core resolution planner that turns a loaded profile into a resolved environment map, dispatching each entry to the right materialization path. This is pure core logic over abstractions core already owns — it does **not** spawn any process.

## What to build

`resolve_profile(profile, opts) -> ResolvedEnv` in `vaultkeeper-core`. For each entry:

- `materialize: "secret"` → retrieve the real value via `SecretBackend::retrieve`.
- `materialize: "lease"` → mint a VaultKeeper lease (JWE) via the existing setup/mint path.

The result is a plain env map (var name → resolved string), a value core can unit-test without spawning anything.

## Hard constraints

- **Written against the `SecretBackend` trait object only.** The resolver must **not** name a concrete backend type. Building against the trait means rung-2 resolution works today on core's `file` backend and works over every JS-implemented backend the moment that backend becomes registry-reachable. This is the smallest possible dependency and is load-bearing for the MCP wrapper use case.
- **All-or-nothing failure semantics.** If any entry fails to resolve, the whole operation aborts **before anything reaches a launcher**. There must be no partial env map and **zero resolved values on any error path** — a failure of one entry must not leave other entries' secrets sitting resolved in memory or returned to a caller.
- **Does not spawn.** Process launch is a separate, per-host concern (the `run` verb). Core resolves to a value and stops there. The resolver must never route through `HostPlatform::exec` (that API captures output, which is wrong for this path).
- Policy defaults come from `VaultConfig::defaults`; per-entry policy overrides them.

## Acceptance criteria

Each numbered criterion names its test layer.

1. **(unit)** A mixed-rung profile resolves so that a `materialize:"secret"` entry yields the real secret value in its env var and a `materialize:"lease"` entry yields a **decryptable JWE lease** (decrypt it in the test and confirm its claims) in its env var — both present in one resolved map.
2. **(unit)** Given a 4-entry profile where the 3rd entry fails to resolve, the call returns an error and produces **no env map at all** — assert that none of the other three entries' resolved values are observable on the error path.
3. **(unit)** The resolver is exercised against `InMemoryBackend` with **no process involved**, demonstrating it is decoupled from any launcher and from any concrete backend type.

## Proof

`cargo test` green including the negative/all-or-nothing cases; `pnpm check` and full TS tests green; rebuild the WASM module since Rust is touched (committed output).

## Complexity

[...]

Issue #277, opened 2026-07-21
Add environment profile schema, loader, and profile subcommands to vaultkeeper-core


mike-north:
Refs #276

Introduce the **environment profile** primitive in `vaultkeeper-core`: the serde schema, a fail-closed loader, and the `profile init/show/list/lint` subcommands. All schema, validation, and output rendering live in core; only argument parsing is per-host.

## Background

A profile is a named, declarative binding set: env-var name → secret source → materialization mode → policy. It is the composition primitive the codebase lacks today (`exec` redeems exactly one JWE under a hardcoded name). Profiles live at `$CONFIG_DIR/profiles/<name>.json` (plus `--profile-file <path>` for a project-local file) — **not** inside `config.json`, which is machine-global vault machinery with a different lifecycle. A profile contains no secrets, only names and policy, so it is committable; that makes policy loosening a reviewable diff.

## Schema

Serde types live in core with `rename_all = "camelCase"` and `deny_unknown_fields`. This exact example must round-trip:

```jsonc
{
  "version": 1,
  "name": "github-mcp",
  "entries": {
    // Rung 2 — unmodified third-party server reads a real token.
    "GITHUB_TOKEN": {
      "secret": "github-pat",
      "materialize": "secret",
      "minTrust": "registry",
      "requirePresencePerUse": false
    },
    // Rung 3 — vaultkeeper-aware consumer; env carries a secret-access lease.
    "VK_DB_CREDENTIAL": {
      "secret": "prod-db-password",
      "materialize": "lease",
      "minTrust": "sigstore",
      "ttlSeconds": 900,
      "useLimit": 5
    },
    // Rung 3 — session signing lease. NOT unattended-restart-safe.
    "VK_SIGNING_LEASE": {
      "signingKey": "release-signer",
      "materialize": "lease",
      "ttlSeconds": 28800,
      "useLimit": 200,
      "requirePresenceAtMint": true
    }
  }
}
```

Entry source is a **discriminated union — exactly one of `secret` or `signingKey`.**

### `materialize` is string-or-object (polymorphic by design; v1 = strings only)

Serde shape: an **untagged enum** `Materialize::Simple(MaterializeMode)` (implemented) | `Materialize::Extended { mode, .. }` (parsed, then refused at validation).

- `"secret"` — resolve to the real value.
- `"lease"` — resolve to a VaultKeeper lease (JWE).
[...]

Issue #276, opened 2026-07-21
Epic: environment profiles and vaultkeeper run (policy-enforced env injection)


mike-north:
## Summary

vaultkeeper today can redeem exactly one secret at a time: `exec --token <jwe> -- cmd` injects a single JWE under the hardcoded name `VAULTKEEPER_SECRET`. This epic introduces the composition primitive that is missing — a **named environment profile** — plus a new `run` verb that resolves a whole profile into a child process's environment, minting where policy requires it.

### The profile primitive

An **environment profile** is a named, declarative binding set: env-var name → secret source → *materialization mode* → policy. It lives at `$CONFIG_DIR/profiles/<name>.json` (or a project-local `--profile-file <path>`), **not** inside `config.json`. A profile contains no secrets — only names and policy — so it is safe and intended to commit beside `package.json`. The deliberate consequence: **weakening a policy becomes a reviewable diff**, subject to the same approval path as a code change.

### Two materialization modes (per entry, not per profile)

Each entry declares `materialize: "secret" | "lease"`:

- **`"secret"`** — resolve to the real secret value. An unmodified third-party tool reads a real token from its environment.
- **`"lease"`** — resolve to a **VaultKeeper lease**: a serialized, expiring, revocable capability (a JWE) that must be presented back to the issuer to be redeemed — granted for a time, not a copy of the secret.

"lease" is the house term for any such capability. A single resolved environment can hold a plaintext value for an unmodified consumer and a lease for a vaultkeeper-aware consumer **simultaneously** — that mixing is the steady state, not a migration phase.

### The `run` verb

`vaultkeeper run --profile <name> -- <command>...` resolves the named set, mints where needed, and launches the command with the resolved environment. It is a new verb, distinct from `exec` (both persist permanently): `exec` is token redemption ("I hold a capability, spend it"); `run` is environment composition ("resolve this named set, minting where needed, then launch").

### Headline use case: the MCP wrapper

An MCP client config becomes:

```jsonc
{ "command": "vaultkeeper",
  "args": ["run", "--profile", "github-mcp", "--", "npx", "-y", "@github/mcp-server"] }
```

[...]

PR #301, opened 2026-07-22
Add profile resolution planner and materialization engine (vaultkeeper-core)

mike-north:
## Summary

Implements the core profile resolution planner and materialization engine:
`resolve_profile(profile, opts) -> ResolvedEnv` in a new `vaultkeeper-core::resolve`
module. For each `LoadedProfile` entry (issue #277's profile primitive):

- `materialize: "secret"` retrieves the real value via `SecretBackend::retrieve`.
- `materialize: "lease"` mints a VaultKeeper lease (JWE) by building the same
  `VaultClaims` shape and calling `jwe::create_token` — the exact mint primitive
  `VaultKeeper::setup()` uses internally. `resolve_profile` deliberately does not
  call the full `setup()` method: its executable-trust binding (Sigstore/TOFU
  verification of the *calling* executable, which needs a `HostPlatform`) is an
  orthogonal, one-time-per-invocation concern for whichever front end drives
  resolution later (e.g. a future `run` verb), not a per-entry concern of the
  resolver. This keeps `resolve_profile` free of any `HostPlatform` dependency. A
  resolved lease's `exe` claim is bound to the `"dev"` sentinel (unverified),
  exactly as `SetupOptions::skip_trust` would produce.

A `materialize: "lease"` entry backed by a `signingKey` source (a session
signing lease) is out of scope here — the epic (#276) explicitly lists
"session mint, presence-at-mint... (to be filed)" as a separate future work
item — and is refused with a clear `VaultError::Other` naming the entry,
rather than silently mishandled.

## Hard constraints, and how they're met

- **Written against `SecretBackend` only:** `ResolveOptions` holds a
  `&dyn SecretBackend`; the resolver never names a concrete backend type.
- **All-or-nothing failure semantics:** `resolve_profile` resolves every entry
  into a scratch `Vec<(String, Zeroizing<String>)>` first, and only converts it
  into the returned `ResolvedEnv` once every entry has succeeded. Any entry
  failure returns that error immediately via `?`, dropping (and zeroizing) the
  scratch buffer — no partial map is ever returned or observable.
- **Does not spawn:** `resolve_profile`/`ResolveOptions` take no `HostPlatform`
  at all, so there is no way to route through `HostPlatform::exec`.
- **Policy defaults from `VaultConfig::defaults`, per-entry overrides:** already
  resolved by the #277 loader into each `ProfileEntry` (`min_trust`,
  `ttl_seconds`) — `resolve_profile` consumes those already-resolved fields.

1. Mixed-rung profile: a `materialize:"secret"` entry yields the real value,
   and a `materialize:"lease"` entry yields a decryptable JWE lease (decrypted
   and its claims asserted) in the same map — covered by
   `resolve::tests::resolves_a_mixed_rung_profile_to_one_map_with_a_real_secret_and_a_decryptable_lease`.
2. A 4-entry profile where the 3rd entry fails to resolve returns an error with
   no env map at all, and none of the other three entries' values are
   observable — covered by
   `resolve::tests::a_failing_entry_aborts_the_whole_call_with_no_partial_env_map_observable`.
3. Resolver exercised against `InMemoryBackend` with no process/launcher
   involved (no `HostPlatform` in `ResolveOptions` at all) — covered by
   `resolve::tests::resolves_against_in_memory_backend_with_no_process_or_launcher_involved`.

All four tests (including the not-yet-supported signing-key-lease guard) live
in `crates/vaultkeeper-core/src/resolve.rs`.

## Proof

- `cargo test --workspace`: all suites green (243 tests in `vaultkeeper-core`'s
  lib target, including the 4 new `resolve::tests::*`; 0 failures anywhere in
  the workspace).
- `cargo fmt --check`: clean.
- `cargo clippy --target wasm32-unknown-unknown -p vaultkeeper-core -p vaultkeeper-wasm --lib -- -D warnings`: clean.
- WASM rebuilt with the pinned toolchain (rustc 1.95.0 / wasm-pack 0.15.0 /
  wasm-opt 131, matching `crates/vaultkeeper-wasm/wasm-toolchain.env`) and the
  regenerated `vaultkeeper_wasm_bg.wasm` committed (binary content changed;
  `.js`/`.d.ts` unchanged; export/import fingerprint unchanged: 47 exports / 54
  imports both sides). Well under the size budget (514,637 bytes raw vs.
  1,048,576 budget; 200,005 bytes gzip vs. 409,600 budget).
- `pnpm check` and `pnpm test`: green (263 passed, 19 pre-existing skips, 0
  failures).

## Notes

- `Cargo.lock` is also included in this PR: it was stale (pinned crate versions
  at `0.3.0` against a workspace already bumped to `0.4.0` on `main`), and
  running `cargo build`/`cargo test` here regenerated it to match. Unrelated to
  this change's logic, included because leaving the workspace `cargo test`
  unrunnable clean isn't otherwise fixable without it.

Refs #278

### DIFF
--- a/crates/vaultkeeper-core/src/errors.rs
+++ b/crates/vaultkeeper-core/src/errors.rs
@@ -289,16 +289,29 @@ pub enum VaultError {
     },
 
     // --- Environment Profile Failures (issue #277) ---
-    /// A profile's `materialize` field used the reserved object form
-    /// (`{ "mode": "...", ... }`). The object form is polymorphic by design
-    /// but v1 only implements the plain string values (`"secret"` |
-    /// `"lease"`) — every object-form `mode` is refused with this typed
-    /// error rather than a generic parse failure, so the reservation is
-    /// discoverable and a real v2 implementation can be non-breaking later.
+    /// A profile requested a materialization the current implementation
+    /// does not support. Two request shapes produce this error:
+    ///
+    /// 1. The `materialize` field used the reserved object form
+    ///    (`{ "mode": "...", ... }`). The object form is polymorphic by
+    ///    design but v1 only implements the plain string values (`"secret"`
+    ///    | `"lease"`) — every object-form `mode` is refused with this typed
+    ///    error rather than a generic parse failure, so the reservation is
+    ///    discoverable and a real v2 implementation can be non-breaking
+    ///    later. `mode` carries the reserved mode name (e.g. `"reference"`).
+    /// 2. A materialization *combination* the resolver cannot yet satisfy,
+    ///    refused at resolve time. `mode` then carries a stable kebab-case
+    ///    slug naming the unsupported request — currently
+    ///    `"signing-key-lease"` (session signing leases, pending the epic's
+    ///    session-mint work) and `"secret-lease-presence-at-mint"`
+    ///    (`requirePresenceAtMint` on a secret-backed lease; no
+    ///    `HostPlatform` exists at resolve time to prompt with). These slugs
+    ///    are a documented, stable contract for programmatic callers.
     #[error("{message}")]
     MaterializeModeUnsupported {
         message: String,
-        /// The reserved `mode` name the profile requested (e.g. `"reference"`).
+        /// The reserved `mode` name or unsupported-request slug — see the
+        /// variant docs for the enumerated values.
         mode: String,
     },
 
--- a/crates/vaultkeeper-core/src/lib.rs
+++ b/crates/vaultkeeper-core/src/lib.rs
@@ -12,6 +12,7 @@ pub mod identity;
 pub mod jwe;
 pub mod keys;
 pub mod profile;
+pub mod resolve;
 pub mod signing;
 pub mod types;
 pub(crate) mod util;
@@ -24,6 +25,7 @@ pub use backend::{
 };
 pub use errors::{ExecutableTrustRequiredReason, VaultError};
 pub use identity::{HANDLE_TABLE_MAX_SIZE, HandleId, HandleTable, StoredClaims};
+pub use resolve::{ResolveOptions, ResolvedEnv, resolve_profile};
 pub use types::{
     BackendConfig, ClaimsKind, ExecRequest, ExecResult, FetchRequest, KeyStatus, LeasePresence,
     PreflightCheck, PreflightCheckStatus, PreflightResult, ScopedPreflightCheck, SecretAccessor,
--- a/crates/vaultkeeper-core/src/profile/loader.rs
+++ b/crates/vaultkeeper-core/src/profile/loader.rs
@@ -267,6 +267,12 @@ fn validate_entry(
     })
 }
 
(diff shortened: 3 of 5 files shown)


mike-north:
Reviewer adjudication of the open threads (fix batch pushed as 639a4de):

**Fixed in 639a4de**: filesystem doc claim (3626637129), signing-key refusal now typed and entry-named (3626637164), requirePresencePerUse enforced through the shared primitive with before-retrieve spy tests (3626717565), explicit ttlSeconds:0 rejected at load time (3626717576), plus: requirePresenceAtMint on a secret-backed lease refused instead of silently minted presence-less, a ResolveOptions::executable_hash seam so Sigstore/TOFU leases bind a verified identity (ExecutableTrustRequired when absent), a shared claims-builder deduping setup()/mint paths, and coverage for zero-entry profiles, key-manager failure, default-TTL flow, and the minTrust matrix.

**Already fixed in 818d24b** (prior round): TTL None-fallback (3626637152), lease-expiry overflow, and the de-tautologized AC2 short-circuit test.

**Deferred (pre-existing convention, tracked)**: zeroizing the retrieved-secret String on the lease path — shared with setup(), not a regression here.

Note: an unsubmitted pending review on this PR currently blocks API replies/resolution on its threads — dispositions recorded here; threads get resolved once the pending review is submitted or withdrawn.

mike-north:
Second-wave adjudication (fix pushed as 259ec33): the retrieved plaintext in resolve_entry is now held in a Zeroizing wrapper so the lease branch scrubs it on exit, success or error (3631959631); the secret branch hands ownership out at the last moment, matching the module's stated zeroization discipline.

mike-north:
Thread-resolution record (replies were blocked by the pending review, so dispositions are logged here; every item was verified fixed at the current head by an adversarial re-audit): filesystem-doc claim corrected; explicit ttlSeconds:0 rejected at load with defensive resolver fallback; signing-key-lease refusal typed and entry-named; requirePresencePerUse enforced via the shared primitive with before-retrieve spy tests; lease-path plaintext held in Zeroizing; and the two MaterializeModeUnsupported taxonomy threads are addressed by widening the variant's documented contract to enumerate the stable unsupported-request slugs (signing-key-lease, secret-lease-presence-at-mint) — a dedicated variant was weighed and declined to keep the #251 wire-code table stable.

mike-north:
Latest thread (build_and_mint_claims leaving its claims.val copy unzeroized): confirmed and fixed — the helper now scrubs its val copy immediately after create_token returns, success or error, so no second unzeroized heap copy outlives the mint. (Resolved without an inline reply — the pending review still blocks those.)

mike-north:
Latest three threads, all fixed: the scratch-map conversion now moves values out of Zeroizing (no extra plaintext clone); the private intra-doc link is plain text via the public resolve_profile reference; and the silent 30-day TTL cap became a typed load-time ConfigValidation (tested at the boundary both sides) with the resolver's cap retained as documented defense in depth.

Issue #290, opened 2026-07-21
Port KeychainBackend to vaultkeeper-core (security -i command-stream store)


mike-north:
Refs #288

Port the macOS Keychain backend (`KeychainBackend`) into `vaultkeeper-core` as a subprocess-orchestration backend driving `security(1)` over `HostPlatform::exec`. This is the flagship backend for the product's positioning ("use your free macOS Keychain instead of a paid tier"), so it must land as a clean, uniform subprocess implementation — no native FFI, no host split.

## The `store` argv-leak problem (and its resolution)

The TS backend's `store()` passes the base64 secret as `security add-generic-password … -w <encoded>` — the secret lands in the `security` process's **argv**, visible to `ps`/Activity Monitor and to any process that can read process arguments. That is the same class of durable exposure tracked for the Keychain backend today. And `security`'s **direct** `-w` invocation has no stdin channel: `-w <value>` puts the value in argv, and `-w` with no value falls back to an interactive `getpass()` TTY prompt that ignores piped stdin.

**Resolution — `security` interactive mode (`security -i`).** In interactive mode, `security` "will enter interactive mode and allow the user to enter multiple commands on stdin. When EOF is read from stdin security will exit." The OS-level argv of the process is then just `security -i`; the entire command — including `add-generic-password … -w <value>` — rides the **stdin command stream**, not the process argv. `ps`/Activity Monitor see only `security -i`. The secret never enters any OS argv, and the `store` path stays a plain `HostPlatform::exec` call (which already carries stdin). Do **not** use native `Security.framework` FFI (not wasm-portable; would force a host split) or accept a TTY-only limitation.

**base64 is load-bearing — keep it.** The value is base64-encoded first (`Buffer.from(secret).toString('base64')`, preserved in the port). The base64 alphabet (`[A-Za-z0-9+/=]`) contains no character `security`'s tokenizer treats specially, so `-w <base64>` is always exactly one clean token on the command stream — this is what neutralizes the space/quote/newline/non-ASCII tokenization hazard. Keep the base64 step; it is load-bearing, not incidental.

## Required pre-implementation verification (one command, no live keychain data)

Confirm that `add-generic-password … -w <value>` parses its inline value from the `-i` stdin stream rather than diverting to the `getpass` prompt (the man page makes this the expected behavior — the `getpass` fallback is specifically the *no-value* trailing-`-w` case — but it is load-bearing, so verify before building on it):

```
[...]

Issue #270, opened 2026-07-21
KeychainBackend passes the base64-encoded secret in the security(1) command line


mike-north:
## Summary

While auditing all `execCommand`/`execCommandFull` call sites for the same
plaintext-in-argv pattern as #269 (DpapiBackend), `KeychainBackend.store()`
(`packages/vaultkeeper/src/backend/keychain-backend.ts`) was found to have the
same class of exposure:

```ts
async store(id: string, secret: string): Promise<void> {
  const service = `${SERVICE_PREFIX}${id}`
  const encoded = Buffer.from(secret, 'utf8').toString('base64')
  await execCommandFull('security', ['delete-generic-password', '-a', ACCOUNT, '-s', service])
  await execCommand('security', [
    'add-generic-password',
    '-a',
    ACCOUNT,
    '-s',
    service,
    '-w',
    encoded,
  ])
}
```

The base64-encoded secret is passed as the `-w` argument, which lands in the
`security` process's argv — visible to any other local process that can list
this process's command line (`ps`, Activity Monitor, `/proc`-equivalents).

## Why this was not fixed in #269

Confirmed by live testing on macOS: the `security` CLI's
`add-generic-password -w` has no non-interactive stdin channel.

- `security add-generic-password -w <value>` puts the password directly in argv (the current behavior).
- `security add-generic-password ... -w` with `-w` as the last flag and no value switches to an *interactive* prompt (`password data for new item:` / `retype password for new item:`) that reads from the controlling TTY via `getpass()`, not from a piped stdin — piping a value on stdin in that mode does not supply the password (verified: piped input is ignored and the prompt still blocks/fails non-interactively).
- There is no documented flag to read the password from a file or stdin.

So, unlike `DpapiBackend` (fixed in #269 by switching to a stdin-fed
PowerShell script) or `SecretToolBackend` (already uses `execCommand(..., { stdin: secret })`,
matching `secret-tool store`'s documented stdin support), there is no
argv-avoiding channel available through the `security` CLI as currently used.

## Possible directions (not decided here)

- Bind directly to the macOS Keychain Services C API (e.g. via a native
  addon or FFI) instead of shelling out to `security`, eliminating the argv
  exposure entirely. This is a larger change and overlaps with the
  consolidation direction discussed in #245 ("port keychain/dpapi/secret-tool/yubikey
  backends into the core as subprocess-wrapper backends").
- Accept the interactive-TTY-only non-exposure and document the current
[...]

PR #304, opened 2026-07-22
Port KeychainBackend to vaultkeeper-core over security -i stdin

mike-north:
Refs #290

## Pre-implementation verification (run FIRST, per issue instructions)

```
$ printf 'add-generic-password -a t -s vk-spike -w dGVzdA==\n' | security -i
$ security find-generic-password -a t -s vk-spike -w
dGVzdA==
```

Exit code 0, no getpass prompt, no TTY hang. The value round-tripped byte-for-byte through
`find-generic-password -w`, confirming `add-generic-password ... -w <value>` is parsed off the
`-i` stdin command stream rather than diverted to the interactive `getpass()` prompt. Verified the
inverse too: a bare trailing `-w` (no inline value) *does* trigger `password data for new item:
retype password for new item:` prompt text even inside `-i` mode — confirming `-w <value>` must
always be inline, never bare.

Design proceeded on this confirmation.

## What this does

Ports the macOS `KeychainBackend` (`packages/vaultkeeper/src/backend/keychain-backend.ts`) into
`vaultkeeper-core` as `crates/vaultkeeper-core/src/backend/keychain.rs` — a pure
`HostPlatform::exec` subprocess orchestrator over `security(1)`, no Security.framework FFI, no
host split. Same account (`vaultkeeper`)/service (`vaultkeeper:<id>`) naming as the TS backend, so
entries are mutually readable.

**`store`'s argv-leak fix:** the whole `delete-generic-password` (pre-clean) +
`add-generic-password ... -w <base64>` command pair is sent as a two-line script over a single
`security -i` process's stdin. The OS-visible argv for that process is just `security -i` — the
secret and its base64 encoding never appear in any child process's argv. `security -i`'s overall
exit code reflects the *last* command in the stream, so a failing pre-delete (nothing to delete)
followed by a successful add still yields exit 0 (verified empirically before relying on it).
base64-encoding is kept unchanged from the TS backend — it's what keeps `-w <value>` a single
clean token for `security -i`'s own stdin tokenizer regardless of the secret's byte content.

**Scope, per the issue:** only `store` uses the `-i` stream. `retrieve`
(`find-generic-password -w`, secret on stdout only), `delete`/`exists` (no secret in argv at all),
and `list` (`dump-keychain` + parse) are plain argv `security <args>` calls, ported 1:1.

**Account/service quoting in the `-i` stream:** unlike a plain argv call, the `-i` stdin stream is
tokenized by `security` itself. The account and `vaultkeeper:<id>` service are embedded as
double-quoted, backslash-escaped tokens — verified empirically to round-trip service names with
spaces, embedded double quotes, and embedded backslashes. (An embedded newline in `id` would break
the single-line `-i` command and is out of scope, consistent with `id` being a single-line
identifier everywhere else in vaultkeeper.)

**`list` parsing:** generalizes the TS backend's `/0x00000007 <blob>="vaultkeeper:([^"]+)"/g`
regex to also handle `security dump-keychain`'s hex+octal-escaped display form, which it switches
to for any service name containing a backslash or a non-printable-ASCII byte (verified
empirically) — the hex digits are parsed as the authoritative raw bytes rather than relying on the
lossy octal-escaped text.

**`is_available` deviation from the TS backend:** the TS backend probes with `security version`,
which is not a real `security` subcommand — verified it always fails with `unknown command
"version"`, so the TS backend's availability check is unconditionally broken on real `security`.
The Rust port instead probes with `security list-keychains`, a real, read-only, side-effect-free
command, so `is_available()` actually reflects reality. This is an internal probe-command choice,
not a change to any wire-visible behavior (storage format, argv shape, or secret handling) — filed
as a known deviation rather than silently perpetuating a broken check.

1. **(unit) `store` argv-safety** — `store_issues_a_single_security_interactive_process`,
   `store_passes_secret_and_its_base64_encoding_on_stdin_not_argv`,
   `store_argv_sentinel_survives_full_round_trip` (`crates/vaultkeeper-core/src/backend/keychain.rs`).
   A sentinel secret is injected and asserted absent from every recorded child argv across the
   whole store→retrieve→exists→delete flow, with the base64-encoded form checked too.
2. **(unit) byte-for-byte round trip** —
   `round_trips_secret_with_spaces_quotes_and_embedded_newlines`,
   `round_trips_non_ascii_secret_byte_for_byte`.
3. **(unit) retrieve/delete/exists/list are plain exec, list parses awkward names** —
   `retrieve_delete_exists_are_plain_argv_exec_calls` (asserts no `-i` and no stdin on those
   calls), `list_parses_plain_and_hex_escaped_dump_keychain_entries`,
   `parse_dump_ids_matches_ts_backend_output_shape`,
   `parse_dump_ids_ignores_entries_without_the_vaultkeeper_prefix`,
   `round_trips_id_containing_spaces_and_quotes` (id, not just secret, survives the `-i`
   tokenizer).
4. **(unit, negative) typed errors, never garbage** — `retrieve_missing_returns_secret_not_found`,
   `delete_missing_returns_secret_not_found`, `store_failure_surfaces_typed_error_not_panic`,
   `retrieve_surfaces_typed_error_on_malformed_base64_not_garbage_secret`.
5. **Conformance on macOS via ephemeral create-keychain** —
   `crates/vaultkeeper-core/tests/keychain_conformance.rs`
   (`keychain_backend_conformance_against_real_security_binary`), sharing its id/secret fixture
   with `vaultkeeper-conformance`'s `"store succeeds with valid secret"` case. Creates a
   throwaway keychain via `security create-keychain`, makes it the default keychain for the
   test's duration (an RAII guard restores the previous default and deletes the ephemeral
   keychain even on panic), runs the full store→exists→retrieve→list→delete→re-retrieve flow
   against the real binary, then tears down. Gated to skip cleanly off-macOS. New
   `keychain-conformance` CI job on `macos-latest` runs it — **there was previously no macOS CI
   runner in this repo at all**, so this PR adds the first one, alongside the existing Linux-only
   `secret_tool_conformance.rs`'s equivalent gate (which stays ungated in CI per its own header
   comment, since it also needs a D-Bus session this repo's CI doesn't set up yet).

## Coordination with #270

#270 tracks the TS `KeychainBackend`'s equivalent argv leak. This PR does **not** touch the TS
implementation — it only lands the Rust-core port with the `security -i` design the issue
prescribes. #270 stays open for the TS-side fix.

## Test plan

- `cargo test --workspace`: 434 tests passed, 0 failed (261 in `vaultkeeper-core`'s lib including
  22 new keychain unit tests; 1 in the new `keychain_conformance` integration test, run for real
  against this macOS dev machine's `security` binary via an ephemeral keychain — verified cleanup
  restored the prior default keychain and left no leftover keychain files).
- `cargo clippy -p vaultkeeper-wasm --target wasm32-unknown-unknown -- -D warnings`: clean.
- `cargo fmt --check`: clean.
- Rebuilt the wasm artifact with the pinned toolchain (`crates/vaultkeeper-wasm/wasm-toolchain.env`:
  rustc 1.95.0, wasm-pack 0.15.0, binaryen 131 — all already installed locally, matching pinned
  versions) since `vaultkeeper-core` (which the wasm crate links) changed. Export/import
  fingerprint against the previously-committed artifact matched (47 exports / 54 imports, both
  sides) — expected, since `KeychainBackend` isn't wired into any wasm-exported binding — but the
  compiled `.wasm` binary itself differs, so the regenerated artifact is committed per the
  package's regeneration policy. `.js`/`.d.ts` outputs were byte-identical (no interface change).
  Both budgets pass (uncompressed 514,629 / 1,048,576 bytes; gzip 199,485 / 409,600 bytes).
- `pnpm check`: clean (pre-existing `security/detect-object-injection` warnings in
  `packages/cli/src/{config-dir,suggest}.ts` are unrelated to this change).
- `pnpm check:api-docs`: up to date (no public TS API surface changed by this PR).
- `pnpm test`: 263 passed, 19 skipped (expected platform-conditional skips), 0 failed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,6 +41,25 @@ jobs:
           path: target/release/vaultkeeper
           retention-days: 1
 
+  keychain-conformance:
+    name: macOS Keychain Backend Conformance
+    # Runs `KeychainBackend` against the real `security(1)` binary and a real
+    # Keychain, via an ephemeral `create-keychain` (issue #290 AC5). The
+    # `rust` job above only runs on ubuntu-latest, where this test detects
+    # the missing macOS prerequisite and skips itself (see
+    # crates/vaultkeeper-core/tests/keychain_conformance.rs), so it needs its
+    # own macOS runner to actually execute.
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: dtolnay/rust-toolchain@stable
+
+      - uses: Swatinem/rust-cache@v2
+
+      - name: Run keychain conformance test
+        run: cargo test -p vaultkeeper-core --test keychain_conformance -- --nocapture
+
   wasm-guards:
     name: WASM Build Guards (drift + size)
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,6 +50,11 @@ jobs:
     # crates/vaultkeeper-core/tests/keychain_conformance.rs), so it needs its
     # own macOS runner to actually execute.
     runs-on: macos-latest
+    # macOS runner minutes bill at ~10x Linux minutes — cap this well below
+    # the 6h default so a hung interactive `security` process (e.g. an
+    # unexpected TTY prompt) can't run away with cost instead of failing
+    # fast.
+    timeout-minutes: 15
     steps:
       - uses: actions/checkout@v4
 
--- a/crates/vaultkeeper-core/src/backend/keychain.rs
+++ b/crates/vaultkeeper-core/src/backend/keychain.rs
@@ -291,9 +291,18 @@ impl SecretBackend for KeychainBackend {
         // in the stream, so a failing pre-delete (no prior entry) followed
         // by a successful add still yields exit code 0 — verified
         // empirically (see module docs).
+        // `encoded` is quoted so an empty secret still yields a value token
+        // (`-w ""`): a bare trailing `-w` would flip `security` into its
+        // interactive getpass prompt and hang a non-interactive caller.
+        // Base64 output never contains `"`/`\`/newline, so plain quoting is
+        // exact (no escaping needed) — enforced by the debug_assert below.
+        debug_assert!(
+            !encoded.contains(['"', '\\', '\n', '\r']),
+            "base64 output must be quote-safe"
+        );
         let script = format!(
             "delete-generic-password -a {account_token} -s {service_token}\n\
-             add-generic-password -a {account_token} -s {service_token} -w {encoded}\n"
+             add-generic-password -a {account_token} -s {service_token} -w \"{encoded}\"\n"
         );
 
         let output = self
@@ -322,6 +331,17 @@ impl SecretBackend for KeychainBackend {
             // entirely removes any path for a future `security` version (or
             // an unanticipated error mode) to leak the secret or its base64
             // encoding into a `VaultError` message that might be logged.
+            // Locked-keychain detection mirrors retrieve/delete/exists —
+            // still without embedding stderr (the signature check reads it,
+            // the error message never includes it).
+            if Self::is_locked_keychain_error(&output.stderr) {
+                return Err(VaultError::BackendLocked {
+                    message: format!(
+                        "macOS Keychain is locked and cannot be written non-interactively for {id}"
+                    ),
+                    interactive: true,
+                });
+            }
             return Err(VaultError::Exec {
                 message: format!(
                     "security add-generic-password failed with exit code {}",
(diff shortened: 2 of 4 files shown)


mike-north:
Reviewer adjudication of the 14 open threads (fix batch in progress on this branch):

**Already fixed at current head** (verified against code): search-list replacement + docstring (3626656638, 3626674788), RAII Drop restoring both (3626656646), macos job timeout-minutes (3626674784), stderr omitted from store() errors with regression test (3626674793), quoted-path parsing without split_whitespace (3626674812).

**Refuted**: RealMacHost stdin-close concern (3626727729) — Child::wait_with_output() drops stdin before waiting, by documented stdlib design.

**Real — fixes incoming in the next push**: id-newline injection into the -i stream (3626656618; ids will be rejected at the boundary before any exec), quoted-id parse truncation with store/list round-trip coverage (3626674798, 3626727750), locked-keychain detection mapping to BackendLocked instead of SecretNotFound (3626727762, 3626727774), no-timeout comment correction (3626674808), and documentation of the non-atomic overwrite window (3626674805).

Note: several of these threads live in an unsubmitted pending review, which blocks API replies — dispositions recorded here instead; threads will be resolved once the fixes land.

mike-north:
Test-hygiene flag from the maintainer side: the test suite on this branch appears to trigger **interactive macOS keychain prompts** during runs (observed: a GUI unlock request for a test keychain named `vk-locktest2` on a developer machine). Test keychains must be created with `security create-keychain -p <pw>` and unlocked non-interactively (`security unlock-keychain -p <pw>`) so the suite never prompts — a GUI prompt hangs CI and any unattended run, which is exactly the failure mode the presence/unattended design work in this repo keeps guarding against. Please audit every `security` invocation in the tests for prompt-freedom (including the lock/relock cases, where an unlock after `security lock-keychain` must supply the password explicitly) and add whatever teardown removes stray test keychains (`security delete-keychain`) even on assertion failure — a leaked test keychain generating prompts on an end-user machine is how this was noticed.

mike-north:
Second-wave adjudication (fixes pushed as 7e50677): the -w value token is now quoted so an empty secret cannot fall into security's interactive getpass prompt (3632004268, with a mock-tokenizer test); store's -i path maps the locked-keychain signature to BackendLocked like the read paths, still without embedding stderr (3632004293); the ephemeral-keychain harness asserts its session-state capture succeeded before mutating global state so Drop can always restore it (3632004328); and run_security's doc now matches its spawn-only-panic behavior, with a strict run_security_ok used for the setup reads (3632004348). All local gates green including both real-binary conformance tests. Threads will be resolved once the pending review is submitted or withdrawn.

mike-north:
Third-wave item (3632173322) fixed in the latest push: the EphemeralKeychain RAII guard is now constructed before the first global security mutation, so a panic anywhere in the remaining setup still restores the session's search list and default keychain via Drop. Conformance tests re-verified against the real binary.

mike-north:
Thread-resolution record (replies blocked by the pending review; every disposition below was verified fixed at the current head by an adversarial re-audit, including both real-binary conformance tests): multiline/NUL ids rejected before any exec with an injection-proof test; quoted-id list parsing fixed via last-quote delimiting with real dump-keychain-shaped unit tests and a real-binary round trip; the locked-keychain signature maps to BackendLocked on all four operations with stderr never embedded; the -w token is quoted so empty secrets cannot hang on getpass; the ephemeral-keychain harness constructs its restore guard before any global mutation and asserts its session-state capture; and the doc corrections (set-keychain-settings comment, overwrite-window and store/store-race notes, load-bearing hyphen) are in.

Issue #282, opened 2026-07-21
Gate register_signing_handle behind an issuance-side principal check; it currently ships a durable ambient signing capability


mike-north:
## Finding

`VaultKeeper::register_signing_handle(kid, backend_ref, expires_at: Option<u64>)` (crates/vaultkeeper-core/src/vault.rs) is a **public** method that mints a signing-key capability handle which is:

- **repeatable** — `resolve_signing_claims()` is non-consuming; the handle keeps producing signing claims for as long as it is live (unlike `read_secret()`, which is one-time),
- **potentially never-expiring** — `expires_at: None` is accepted, and the expiry check treats `None` as "never expires",
- **redeemable by mere possession** of the opaque `HandleId`, with **no invocation-time principal check** anywhere in the system (there is no executable-hash / trust / presence check at *use* time; the `exe` claim is validated only for non-emptiness at mint time and never re-checked).

Net effect: this converts a per-call signing capability into a **durable ambient signing capability** that any same-process code holding the handle id can redeem. A completed custody survey found that adding an unlimited-use, long-lived signing handle *without* an issuance-side principal check is strictly worse than today's behavior.

## Reachability today

Not reached by the CLI or wasm end-user surfaces (no callers there), and `authorize()` mints only one-time secret handles with finite `expires_at = claims.exp`. But `register_signing_handle` is shipped **public API** on `VaultKeeper`, and the `crate::identity::handles` module docs actively *endorse* the long-lived non-interactive signing case ("a long-lived, machine-driven caller with an unlimited or high `use_limit` keeps working — no human in the loop") as a supported goal rather than an out-of-scope, gated one.

## Required

1. **Gate exposure on an issuance-side principal check.** No long-lived / unlimited / never-expiring signing handle class should be mintable until an invocation-time principal check exists (open product decision #261). Until then, constrain `register_signing_handle`: reject `expires_at: None` for signing handles (require a finite expiry), and consider narrowing its visibility (`pub(crate)` / `#[doc(hidden)]`) so it is not a public affordance.
2. **Reword the module + method docs** from "long-lived non-interactive signing is fully supported/expressible" to "deliberately out of scope for now; gated on the issuance-side principal check tracked in #261." Leave the lifetime model structurally open (don't hard-code short-lived-only assumptions) but do not paint 'walk through here' on the open door.
[...]

PR #318, opened 2026-07-22
Add product brief and PRFAQ

mike-north:
Adds the two durable product documents the repo has lacked: a statement of what vaultkeeper is for, and the reference against which scope and security claims get decided.

**`docs/product/PRFAQ.md`** — press-release framing of the value proposition (policy-enforced secret custody over the credential store you already own, no subscription), plus an FAQ that is deliberately honest about limits: what the tool does and does not defend against, what exists today versus what is in flight, and the non-goals.

**`docs/product/product-brief.md`** — the deciding document: single value proposition, storage-vs-policy positioning, the three-rung adoption gradient (plaintext → resolved environment → lease), personas, core concepts in the house vocabulary (lease, minTrust, presence, profiles), the security-claim policy (a claim appears in docs only if code enforces it), non-goals, current state, and success measures.

Docs-only; no package changes, no changeset. Content reflects the shipped 0.7.x surface and the active epics (profiles/run, signing leases, backend consolidation, test strategy) without internal process references.

### DIFF
--- a/docs/product/PRFAQ.md
+++ b/docs/product/PRFAQ.md
@@ -1,8 +1,11 @@
-# vaultkeeper PRFAQ
+# vaultkeeper 1.0 PRFAQ
 
-> Product reference document, written from the launch perspective: the press release and FAQ
-> describe vaultkeeper as it lands at launch. The "What exists today" question anchors current
-> implementation status against that. Kept current by the maintainers.
+> This is the PRFAQ for **vaultkeeper 1.0 — a release that has not shipped yet.** It is
+> written from 1.0's launch perspective, deliberately without hedging: everything described
+> here is in scope for 1.0 whether or not it has landed today. This document is the working
+> definition of what 1.0 means, and the artifact against which product feedback is given.
+> Current implementation status lives where it stays current — the repository's releases,
+> epics, and issue tracker — never here.
 
 ---
 
@@ -114,10 +117,11 @@ permissions. Every lease begins and ends with this key.
 
 A lease is an encrypted token, **minted** with the minting key, that occupies the env-var slot
 *instead of* a secret. It carries expiry, a usage limit, and revocation state, and it can only
-be **redeemed** — decrypted and honored — by the vaultkeeper that holds the minting key. That
-is why you should care: a lease captured into a log aggregator, a telemetry payload, or a bug
-report is dead on arrival anywhere else, because no other machine's vaultkeeper can redeem it.
-That is the leak class that actually happens, and it simply stops mattering.
+be **redeemed** — decrypted and honored — by the vaultkeeper that holds the minting key, *and
+only when presented by the process it was issued to*. That is why you should care: a lease
+captured into a log aggregator, a telemetry payload, or a bug report is dead on arrival —
+no other machine's vaultkeeper can decrypt it, and even another process on the same machine
+is refused at redemption. The leak classes that actually happen simply stop mattering.
 
 Signing leases are the same idea for signing keys: holding one lets a process *ask vaultkeeper
 to produce signatures* with a key it never sees — minted with human presence at launch, usable
@@ -150,9 +154,9 @@ neighbors stay put.
    than the target system itself can express**. GitHub's permission model cannot say "may
    update issues, but only ones carrying this label"; a delegated fetch through vaultkeeper
    can, because each invocation is evaluated against your policy before the key is ever used.
-   This level requires the consuming tool to speak vaultkeeper; today that means first-party
-   tooling, by choice (a language-agnostic local redemption endpoint is a named future epic,
-   gated on an installed base worth integrating against).
+   This level requires the consuming tool to speak vaultkeeper — natively, through its SDKs,
+   or through the local redemption endpoint that lets any tool in any language trade a lease
+   for what it needs in one call.
 
 ### What is a profile?
 
@@ -222,14 +226,14 @@ Secret-store integrations aren't untestable — they're *annoying and difficult*
 hardware to touch, sessions to authenticate, prompts that hang CI. The idea is that
 vaultkeeper owns those annoying tests so you don't have to write them.
 
-vaultkeeper treats its test doubles as products. Shipped today: an in-memory backend and a
-test-vault harness for hermetic consumer tests. In review as of this writing: that backend
-gaining real Ed25519 signing and scriptable fault injection (throwing the same typed errors
-production throws), plus a presence simulator built to be structurally unreachable from
-production code. Planned behind those: backend-flavored doubles carrying real data shapes and
-choreography, held to the *same conformance corpus* as the real adapters — at which point
-manual testing reduces to a minutes-scale, tool-upgrade-triggered audit: "does the real
-adapter still match its double?"
+vaultkeeper treats its test doubles as products: an in-memory backend with real Ed25519
+signing and scriptable fault injection (throwing the same typed errors production throws), a
+presence simulator that is structurally unreachable from production code, and backend-flavored
+doubles carrying real data shapes and choreography — every double held to the *same
+conformance corpus* as the real adapter it stands in for. Test artifacts are self-announcing
+(a test-minted lease is unmistakably marked, and production refuses marked artifacts), so test
+data can never pass as real. Manual testing reduces to a minutes-scale, tool-upgrade-triggered
+audit: "does the real adapter still match its double?"
 
 ### What are the non-goals?
 
@@ -240,19 +244,14 @@ cert-based auth are also out of scope: users there cannot add servers or edit co
 not have the secret-handling problem vaultkeeper solves. Single machine, single developer,
 policy-enforced: that is the product.
 
-### What exists today, and what is in flight?
+### Is everything here real today?
 
-Shipped: the Rust core (backends, JWE capability tokens, key management and rotation, trust
-tiers/TOFU, presence model, detached-JWS signing with backend-held keys, the opaque handle
-table), the TypeScript SDK and CLI, the WASM bridge, encrypted key-state persistence, the
-environment-profile primitive, and the conformance suite. Policy-resolved injection ships
-today as the single-secret `vaultkeeper exec`; the profile-based `run` wrapper described
-above is its in-flight, multi-secret generalization. Also in flight: session signing leases
-with tamper-evident revocation, native-CLI backend parity (keychain, secret-tool, DPAPI,
-YubiKey ports to the shared core), and the paired-double test strategy. Committed roadmap:
-verifier-visible assurance (presence-bound keys and the signed assurance assertion — see "Can
-a signature prove a human was present?"). Future: the local redemption endpoint, external key
-import with custody provenance.
+No — and that is the point of this document. This PRFAQ describes **vaultkeeper 1.0** and is
+written from its launch perspective: it is the target the project builds toward, and the
+artifact its owner gives product feedback against. Some of what it describes has shipped;
+some is in flight; some is not yet started. The one place to learn which is which is the
+repository itself — releases, epics, and the issue tracker — which stays current the way a
+durable document cannot.
 
 ### What does it cost?
 
--- a/docs/product/PRFAQ.md
+++ b/docs/product/PRFAQ.md
@@ -22,19 +22,26 @@ service tokens, dedicated vaults, per-use biometric approval — largely gate th
 behind paid business tiers or cloud subscriptions.
 
 vaultkeeper unbundles **storage** from **policy**. Your secrets live in whatever backend you
-already trust — macOS Keychain, 1Password, a YubiKey, Windows DPAPI, libsecret, or an encrypted
-file store — and vaultkeeper supplies the governance layer those stores lack: capability tokens
+already trust — today: macOS Keychain, 1Password, a YubiKey, Windows DPAPI, the Linux Secret
+Service (`secret-tool`), or an encrypted file store, with further stores (Dashlane and others)
+on the table wherever demand appears, since a backend is a small adapter over a stable
+interface — and vaultkeeper supplies the governance layer those stores lack: capability tokens
 scoped by expiry, usage count, and executable identity; per-use human-presence requirements
 (a Touch ID tap or YubiKey touch *for this operation, right now*); delegated access patterns
 where the raw secret never enters the calling process at all; and full local revocation.
 
-Adoption is a gradient, not a rewrite. `vaultkeeper run --profile github-mcp -- npx
-@github/mcp-server` wraps any unmodified tool: the plaintext token leaves the config file and is
-materialized only into the child process, under policy, with nothing at rest. Tools that
-integrate further can hold a **VaultKeeper lease** instead of the secret itself — an encrypted,
-expiring, revocable capability that is worthless off the machine and worthless after its window
-closes. Signing keys go further still: the private key never leaves the backend at all;
-consumers request signatures, never key material.
+Adoption is a gradient, not a rewrite, and the wrapper works on anything that reads
+environment variables. Wrap an MCP server so the PAT leaves the config file
+(`vaultkeeper run --profile github-mcp -- npx @github/mcp-server`); wrap an ordinary CLI for
+one policy-resolved invocation (`vaultkeeper run --profile deploy -- terraform apply`); or
+wrap the command that launches an agentic coding session, so the agent's whole environment —
+real secrets and vaultkeeper leases alike — is resolved under policy at launch and nothing
+lives in a dotfile. In every case the tool is unmodified: secrets are materialized only into
+the child process, under policy, with nothing at rest. Tools that integrate further can hold a
+**VaultKeeper lease** instead of the secret itself — an encrypted, expiring, revocable
+capability that is worthless off the machine and worthless after its window closes. Signing
+keys go further still: the private key never leaves the backend at all; consumers request
+signatures, never key material.
 
 The same identity primitives serve human and automation signers alike. An agent fleet can hold
 signing identities whose keys are born inside vaultkeeper and cannot be extracted by the
@@ -69,35 +76,66 @@ floor dramatically without pretending to be a sandbox.
 
 ### How is this different from dotenvx / `op run` / SOPS?
 
-Those tools solve *encrypted at rest, injected at runtime* — and solve it well. vaultkeeper's
-differences are: (1) **policy at resolution** — trust tiers on the calling executable, TTLs,
-usage limits, per-use presence — not just decryption; (2) **storage-agnostic** — the same policy
-over Keychain, 1Password, YubiKey, DPAPI, libsecret, or a file, rather than one blessed store;
-(3) **the lease rung** — consumers can hold a capability instead of the secret, which nothing
-env-injection-shaped offers; (4) **no subscription** — the scoped-token/dedicated-vault posture
-that 1Password gates behind Business tiers, locally and free. What those tools have that
-vaultkeeper deliberately does not: team sync and cloud distribution (see non-goals).
+Those tools solve *encrypted at rest, injected at runtime* — and solve it well. But they are
+not context-aware: they decrypt for whoever invokes them, so in practice you choose between
+security and friction (lock things down and pay for it on every invocation, or loosen them
+and hope). Because vaultkeeper's resolution step evaluates *policy* — who is asking, with
+what trust, for how long — it can offer both at once: strong defaults with low-friction
+day-to-day use. The concrete differences:
+
+1. **Policy at resolution** — trust tiers on the calling executable, TTLs, usage limits,
+   per-use presence — not just decryption.
+2. **Storage-agnostic** — the same policy over Keychain, 1Password, YubiKey, DPAPI, the
+   Linux Secret Service, or a file, rather than one blessed store.
+3. **The lease rung** — consumers can hold a capability instead of the secret, which nothing
+   env-injection-shaped offers.
+4. **No subscription** — the scoped-token/dedicated-vault posture that 1Password gates behind
+   Business tiers, locally and free.
+
+What those tools have that vaultkeeper deliberately does not: team sync and cloud
+distribution (see non-goals).
 
 ### What is a VaultKeeper lease?
 
 An encrypted, serializable capability token that occupies the env-var slot *instead of* a
-secret. It carries expiry, a usage limit, and revocation state; redeeming it requires the local
-vault key, so a lease captured into a log aggregator or bug report is dead on arrival anywhere
-else. Signing leases authorize signature operations for a bounded session — minted with human
-presence at launch, usable non-interactively until expiry, revocable individually or per-key.
+secret. It carries expiry, a usage limit, and revocation state; redeeming it requires the
+local vault key (the encryption key vaultkeeper generates at setup and keeps in your
+configured backend — only the vaultkeeper process on this machine can use it, enforced by the
+backend's own access control plus owner-only file permissions on the encrypted key state), so
+a lease captured into a log aggregator or bug report is dead on arrival anywhere else. Signing
+leases are the same idea for signing keys: holding one lets a process *ask vaultkeeper to
+produce signatures* with a key it never sees — minted with human presence at launch, usable
+non-interactively until expiry, revocable individually or per-key.
 
-### What is the adoption gradient?
+### How do I adopt this without rewriting my tools?
 
-Three rungs, chosen **per secret**, permanently mixable in one profile:
+You don't adopt vaultkeeper wholesale — you upgrade **one secret at a time**, and each secret
+sits at whichever of three levels fits it. Nothing forces a migration: a single profile can
+hold secrets at different levels side by side, indefinitely, so there is no cliff where
+everything must convert at once and no penalty for upgrading one credential while its
+neighbors stay put.
 
-1. **Plaintext on disk** — the status quo being replaced.
-2. **Resolved environment** — the real secret, never at rest, materialized under policy into
-   one process. Works with any unmodified tool. This is the on-ramp, and the MCP-server wrapper
-   is its headline: point `mcp.json` at `vaultkeeper run` and the PAT leaves the file.
-3. **Lease** — the env var carries a capability, not the secret. Requires the consumer to speak
-   vaultkeeper; today that means first-party tooling, by choice (a language-agnostic local
-   redemption endpoint is a named future epic, gated on an installed base worth integrating
-   against).
+1. **Environment variables holding plaintext secrets in your terminal or tool config** (an
+   `export GITHUB_TOKEN=ghp_…` in `.bashrc`, a token pasted into `mcp.json` or a `.env`) —
+   the status quo being replaced. The secret sits on disk in plaintext, readable by anything,
+   forever.
+2. **Resolved at launch, never on disk.** You run the same tool through
+   `vaultkeeper run`, and the secret is fetched from your backend and handed *only to that
+   one process, for that one run*. If you have ever thought "I just need to pass this key to
+   this one command without leaving it lying around in a file" — this is that, made routine.
+   The tool is unmodified; the difference from level 1 is that the secret is no longer
+   floating around on disk for every other process, backup, and sync client to see.
+3. **The process gets a lease, not the secret.** At level 2 the chosen process still holds
+   the real GitHub key. At level 3 it holds a vaultkeeper lease instead: a token that — under
+   the right conditions — lets operations be performed *with* a key the process never has any
+   direct access to. Those "right conditions" are yours to define: a time window, a usage
+   count, and — because vaultkeeper sits in the middle of every call — policy **more granular
+   than the target system itself can express**. GitHub's permission model cannot say "may
+   update issues, but only ones carrying this label"; a delegated fetch through vaultkeeper
+   can, because each invocation is evaluated against your policy before the key is ever used.
+   This level requires the consuming tool to speak vaultkeeper; today that means first-party
+   tooling, by choice (a language-agnostic local redemption endpoint is a named future epic,
+   gated on an installed base worth integrating against).
 
 ### Does this work for AI-agent fleets?
 
@@ -111,11 +149,18 @@ the processes it governs — custody cannot substitute for process isolation, an
 
 ### Can a signature prove a human was present?
 
-Not yet, and not claimed. Presence is enforced at signing time by capable backends and is
-visible to the local caller, but a verifier of the resulting signature cannot currently
-distinguish a presence-backed signature from an automation one. Making assurance
-verifier-visible is a known, deliberately-deferred design question shared with downstream
-attestation tooling.
+Not in what ships today, but the path exists and is on the roadmap. Presence is enforced at
+signing time by capable backends and is visible to the local caller; the resulting signature,
+however, does not yet carry that fact to a verifier. Two mechanisms can close the gap. The
+strongest is **presence-bound keys**: when a key lives in hardware whose signing policy
+requires a physical touch (a YubiKey with touch-required, a Secure-Enclave key gated on
+biometrics) and that policy is attested at enrollment, then *every* valid signature from that
+key is proof of presence by construction — nothing extra to encode, nothing to forge. The
+complementary mechanism is a **vaultkeeper-signed assurance assertion** bound into the signed
+payload, covering backends where the property is enforced by vaultkeeper rather than by the
+hardware itself (with the honest caveat that its trust root is the vault, not the physical
+token). Sequencing verifier-visible assurance is a design question shared with downstream
+attestation tooling, which currently treats the distinction as consumer-side convention.
 
 ### How do humans and automation coexist as signers?
 
@@ -127,12 +172,14 @@ verifying system's policy, not to vaultkeeper.
 
 ### What about testing? Secrets tooling is notoriously untestable.
 
-vaultkeeper ships its own test doubles as products: an in-memory backend with real Ed25519
-signing and scriptable fault injection (throwing the same typed errors production throws), a
-presence simulator that is structurally unreachable from production code, and — in flight —
-backend-flavored doubles carrying real data shapes and choreography, held to the *same
-conformance corpus* as the real adapters. Manual testing reduces to a minutes-scale,
-tool-upgrade-triggered audit: "does the real adapter still match its double?"
+vaultkeeper treats its test doubles as products. Shipped today: an in-memory backend and a
+test-vault harness for hermetic consumer tests. In review as of this writing: that backend
+gaining real Ed25519 signing and scriptable fault injection (throwing the same typed errors
+production throws), plus a presence simulator built to be structurally unreachable from
+production code. Planned behind those: backend-flavored doubles carrying real data shapes and
+choreography, held to the *same conformance corpus* as the real adapters — at which point
+manual testing reduces to a minutes-scale, tool-upgrade-triggered audit: "does the real
+adapter still match its double?"
 
 ### What are the non-goals?
 
--- a/docs/product/PRFAQ.md
+++ b/docs/product/PRFAQ.md
@@ -1,0 +1,160 @@
+# vaultkeeper PRFAQ
+
+> Internal working document. The press release is aspirational-at-launch framing; the FAQ is
+> honest about what exists today versus what is in flight. Kept current by the maintainers.
+
+---
+
+## Press Release
+
+### vaultkeeper brings policy-enforced secret custody to every developer — using the credential store you already own
+
+**Developers can now get the secret-handling posture that previously required a business-tier
+subscription — scoped tokens, expiring grants, per-use approval — locally, offline, and free,
+on top of macOS Keychain, 1Password, a YubiKey, or an encrypted file.**
+
+Today, the standard way to give a tool a secret is still an environment variable holding a
+plaintext credential, sourced from a dotfile or pasted into a config. Every MCP server config
+with a `GITHUB_TOKEN` in it, every `.env` in a backup, every API key in shell history is the
+same story: a long-lived, unscoped credential at rest in plaintext, readable by anything and
+revocable only by rotating it everywhere at once. The tools that fix this properly — scoped
+service tokens, dedicated vaults, per-use biometric approval — largely gate those capabilities
+behind paid business tiers or cloud subscriptions.
+
+vaultkeeper unbundles **storage** from **policy**. Your secrets live in whatever backend you
+already trust — macOS Keychain, 1Password, a YubiKey, Windows DPAPI, libsecret, or an encrypted
+file store — and vaultkeeper supplies the governance layer those stores lack: capability tokens
+scoped by expiry, usage count, and executable identity; per-use human-presence requirements
+(a Touch ID tap or YubiKey touch *for this operation, right now*); delegated access patterns
+where the raw secret never enters the calling process at all; and full local revocation.
+
+Adoption is a gradient, not a rewrite. `vaultkeeper run --profile github-mcp -- npx
+@github/mcp-server` wraps any unmodified tool: the plaintext token leaves the config file and is
+materialized only into the child process, under policy, with nothing at rest. Tools that
+integrate further can hold a **VaultKeeper lease** instead of the secret itself — an encrypted,
+expiring, revocable capability that is worthless off the machine and worthless after its window
+closes. Signing keys go further still: the private key never leaves the backend at all;
+consumers request signatures, never key material.
+
+The same identity primitives serve human and automation signers alike. An agent fleet can hold
+signing identities whose keys are born inside vaultkeeper and cannot be extracted by the
+processes they govern — with human presence proven once at session establishment rather than
+demanded per signature.
+
+vaultkeeper is a single Rust core reachable from a native CLI and from TypeScript via
+WebAssembly, with a data-driven conformance suite holding every implementation to the same
+behavior. It is open source, local-first, and has no cloud component: nothing about your
+secrets ever leaves your machine.
+
+---
+
+## FAQ
+
+### What problem does vaultkeeper actually solve?
+
+Plaintext credentials at rest, and the absence of policy between a secret and the processes that
+use it. The realistic adversaries it defeats are **exfiltration-at-rest** (dotfiles, backups,
+synced folders, committed configs, shell history) and **accidental disclosure** (logs, bug
+reports, telemetry, copy-paste). It also bounds the blast radius of what a process holds: an
+expiring, usage-limited, revocable lease instead of a forever-credential.
+
+### What does it *not* defend against?
+
+An attacker with arbitrary code execution as your own user, on your unlocked machine. A
+same-UID process can read a child's environment on some platforms, attach a debugger, or drive
+vaultkeeper's own APIs. Presence-gated backends (YubiKey touch, Touch ID) are the exception —
+they force a fresh physical action per operation that no same-UID software can fake. We state
+this plainly because a secrets tool that overclaims is worse than none: vaultkeeper raises the
+floor dramatically without pretending to be a sandbox.
+
+### How is this different from dotenvx / `op run` / SOPS?
+
+Those tools solve *encrypted at rest, injected at runtime* — and solve it well. vaultkeeper's
+differences are: (1) **policy at resolution** — trust tiers on the calling executable, TTLs,
+usage limits, per-use presence — not just decryption; (2) **storage-agnostic** — the same policy
+over Keychain, 1Password, YubiKey, DPAPI, libsecret, or a file, rather than one blessed store;
+(3) **the lease rung** — consumers can hold a capability instead of the secret, which nothing
+env-injection-shaped offers; (4) **no subscription** — the scoped-token/dedicated-vault posture
+that 1Password gates behind Business tiers, locally and free. What those tools have that
+vaultkeeper deliberately does not: team sync and cloud distribution (see non-goals).
+
+### What is a VaultKeeper lease?
+
+An encrypted, serializable capability token that occupies the env-var slot *instead of* a
+secret. It carries expiry, a usage limit, and revocation state; redeeming it requires the local
+vault key, so a lease captured into a log aggregator or bug report is dead on arrival anywhere
+else. Signing leases authorize signature operations for a bounded session — minted with human
+presence at launch, usable non-interactively until expiry, revocable individually or per-key.
+
+### What is the adoption gradient?
+
+Three rungs, chosen **per secret**, permanently mixable in one profile:
+
+1. **Plaintext on disk** — the status quo being replaced.
+2. **Resolved environment** — the real secret, never at rest, materialized under policy into
+   one process. Works with any unmodified tool. This is the on-ramp, and the MCP-server wrapper
+   is its headline: point `mcp.json` at `vaultkeeper run` and the PAT leaves the file.
+3. **Lease** — the env var carries a capability, not the secret. Requires the consumer to speak
+   vaultkeeper; today that means first-party tooling, by choice (a language-agnostic local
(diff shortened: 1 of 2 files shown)


docs/product/PRFAQ.md
```diff
@@ -0,0 +1,160 @@
...
+
+---
+
+## Press Release
+
+### vaultkeeper brings policy-enforced secret custody to every developer — using the credential store you already own
+
+**Developers can now get the secret-handling posture that previously required a business-tier
+subscription — scoped tokens, expiring grants, per-use approval — locally, offline, and free,
+on top of macOS Keychain, 1Password, a YubiKey, or an encrypted file.**
```
mike-north: I don't think there's any reason why we couldn't integrate things like Dashlane or other secrets backends. So let's strike the balance between stating the things that are currently supported but an intent to support other backends where we see demand.

docs/product/PRFAQ.md
```diff
@@ -0,0 +1,160 @@
...
+
+Today, the standard way to give a tool a secret is still an environment variable holding a
+plaintext credential, sourced from a dotfile or pasted into a config. Every MCP server config
+with a `GITHUB_TOKEN` in it, every `.env` in a backup, every API key in shell history is the
+same story: a long-lived, unscoped credential at rest in plaintext, readable by anything and
+revocable only by rotating it everywhere at once. The tools that fix this properly — scoped
+service tokens, dedicated vaults, per-use biometric approval — largely gate those capabilities
+behind paid business tiers or cloud subscriptions.
```
mike-north: The lack of context awareness of these alternatives also means that in the current state of the world, there's a trade-off between security and friction. VaultKeeper by allowing the user to define policy provides a path where you can have security and reduce friction.

docs/product/PRFAQ.md
```diff
@@ -0,0 +1,160 @@
...
+
+vaultkeeper unbundles **storage** from **policy**. Your secrets live in whatever backend you
+already trust — macOS Keychain, 1Password, a YubiKey, Windows DPAPI, libsecret, or an encrypted
+file store — and vaultkeeper supplies the governance layer those stores lack: capability tokens
+scoped by expiry, usage count, and executable identity; per-use human-presence requirements
+(a Touch ID tap or YubiKey touch *for this operation, right now*); delegated access patterns
+where the raw secret never enters the calling process at all; and full local revocation.
+
+Adoption is a gradient, not a rewrite. `vaultkeeper run --profile github-mcp -- npx
```
mike-north: I like that we're showing an example command here, but let's demonstrate the flexibility by showing how this could be used to create a CLI wrapper for a conventional CLI command. Or this example, which is an MCP wrapper. Or even making either secrets or vaultkeeper leases available to an agentic chat session by wrapping the CLI that kicks off that chat session.

docs/product/PRFAQ.md
```diff
@@ -0,0 +1,160 @@
...
+### How is this different from dotenvx / `op run` / SOPS?
+
+Those tools solve *encrypted at rest, injected at runtime* — and solve it well. vaultkeeper's
+differences are: (1) **policy at resolution** — trust tiers on the calling executable, TTLs,
+usage limits, per-use presence — not just decryption; (2) **storage-agnostic** — the same policy
+over Keychain, 1Password, YubiKey, DPAPI, libsecret, or a file, rather than one blessed store;
+(3) **the lease rung** — consumers can hold a capability instead of the secret, which nothing
+env-injection-shaped offers; (4) **no subscription** — the scoped-token/dedicated-vault posture
+that 1Password gates behind Business tiers, locally and free. What those tools have that
+vaultkeeper deliberately does not: team sync and cloud distribution (see non-goals).
```
mike-north: Format these as numbered bullets

docs/product/PRFAQ.md
```diff
@@ -0,0 +1,160 @@
...
+
+### What is a VaultKeeper lease?
+
+An encrypted, serializable capability token that occupies the env-var slot *instead of* a
+secret. It carries expiry, a usage limit, and revocation state; redeeming it requires the local
+vault key, so a lease captured into a log aggregator or bug report is dead on arrival anywhere
```
mike-north: > redeeming it requires the local vault key

this needs a parenthetical to describe what we mean by local vault key, including which pieces of software have access to it and how that is enforced.

docs/product/PRFAQ.md
```diff
@@ -0,0 +1,160 @@
...
+else. Signing leases authorize signature operations for a bounded session — minted with human
```
mike-north: > signature operations

i'm not sure what you mean by signature operations.

docs/product/PRFAQ.md
```diff
@@ -0,0 +1,160 @@
...
+presence at launch, usable non-interactively until expiry, revocable individually or per-key.
+
+### What is the adoption gradient?
```
mike-north: This needs a lot more explanation. It only makes sense to you and me because we've been talking about an adoption gradient, but you have to write this doc in a way that's intended to be read by those not familiar with the project or our discussions about it.

docs/product/PRFAQ.md
```diff
@@ -0,0 +1,160 @@
...
+
+Three rungs, chosen **per secret**, permanently mixable in one profile:
```
mike-north: > permanently mixable 

this is not going to have meaning to a reader unless we connect it with what the consequence of being permanently mixable means.

docs/product/PRFAQ.md
```diff
@@ -0,0 +1,160 @@
...
+
+1. **Plaintext on disk** — the status quo being replaced.
```
mike-north: > **Plaintext on disk**

let's get more specific here and call this something like "environment variables holding plaintext secrets in your terminal config (e.g. `.bashrc`)"

docs/product/PRFAQ.md
```diff
@@ -0,0 +1,160 @@
...
+2. **Resolved environment** — the real secret, never at rest, materialized under policy into
+   one process. Works with any unmodified tool. This is the on-ramp, and the MCP-server wrapper
+   is its headline: point `mcp.json` at `vaultkeeper run` and the PAT leaves the file.
```
mike-north: We need to explain the resolved environment piece in a way that will be relatable to anyone who has managed secrets exposed to commands that are run in a terminal. Maybe the thing that will resonate here is securely passing an environment variable with a secret in it to an individual CLI command.

The key difference between this an "plaintext on disk" is it's no longer floating around on disk for random other things to see it.

docs/product/PRFAQ.md
```diff
@@ -0,0 +1,160 @@
...
+3. **Lease** — the env var carries a capability, not the secret. Requires the consumer to speak
+   vaultkeeper; today that means first-party tooling, by choice (a language-agnostic local
+   redemption endpoint is a named future epic, gated on an installed base worth integrating
+   against).
```
mike-north: This is another one where we could do better by explaining how this third tier improves upon the second. So just like in the second tier, it's only the processes that you choose that get the secret. But with this third tier, you get the policy layer. So you're no longer just passing a GitHub API key to a process. you are passing something that under the right conditions allows operations to be performed using a GitHub API key that the process doesn't have any direct access to. And this means you can define a time window or a boundary of what is and isn't allowed. A key thing that might resonate with readers here is you actually have the ability to define policy that may be more granular than what the permissions system on the system you're calling into allows. For example, permitting GitHub issues to be updated but only if they have a certain GitHub issue label on them. GitHub's permissioning system does not allow this, but when you're doing delegated fetch and you have a policy layer, you have the ability to define on a per invocation basis whether something should or shouldn't be allowed.

docs/product/PRFAQ.md
```diff
@@ -0,0 +1,160 @@
...
+inside a backend and never exported; signatures produced backend-side; session-scoped signing
+leases minted at harness launch with presence proven once; and honest documentation that
+*non-invocability* additionally requires OS-level isolation between the privileged harness and
+the processes it governs — custody cannot substitute for process isolation, and we say so.
+
+### Can a signature prove a human was present?
+
+Not yet, and not claimed. Presence is enforced at signing time by capable backends and is
+visible to the local caller, but a verifier of the resulting signature cannot currently
+distinguish a presence-backed signature from an automation one. Making assurance
```
mike-north: > Presence is enforced at signing time by capable backends and is
visible to the local caller, but a verifier of the resulting signature cannot currently distinguish a presence-backed signature from an automation one.

Is this strictly true? We're the ones that are configuring how we engage with these secret backends, and there are certain mechanisms like a YubiKey where at the moment a signature is applied, we know positively that the human presence was an element of what just happened. why not encode that into the signature in some way (has to be tamper-proof, but I think it's possible)?

mike-north (commented):
Just reviewed the prfaq for now. Use my feedback to also consider whether the product brief needs updates. I will review it later.

docs/product/PRFAQ.md
mike-north: Reworded — backend lists now state today's set plus openness to further stores (Dashlane named) where demand appears, in both docs.

docs/product/PRFAQ.md
mike-north: Added — the comparison now leads with the context-awareness point: without policy at resolution you trade security against friction; policy dissolves the tradeoff.

docs/product/PRFAQ.md
mike-north: Expanded to three wrapper examples: MCP server, ordinary CLI invocation (terraform apply), and wrapping an agentic chat session's launcher so its whole environment resolves under policy.

docs/product/PRFAQ.md
mike-north: Done — numbered list.

docs/product/PRFAQ.md
mike-north: Added the parenthetical: the vault key is the encryption key generated at setup, held in the configured backend, usable only by the local vaultkeeper process — enforced by the backend's access control plus owner-only permissions on the encrypted key state.

docs/product/PRFAQ.md
mike-north: Reworded — a signing lease lets a process ask vaultkeeper to produce signatures with a key it never sees; 'signature operations' is gone.

docs/product/PRFAQ.md
mike-north: Rewrote the whole answer for a reader with zero project context — retitled to 'How do I adopt this without rewriting my tools?', leading with per-secret upgrades and no forced migration before the three levels.

docs/product/PRFAQ.md
mike-north: Connected to its consequence: no migration cliff, no penalty for upgrading one credential while its neighbors stay put — a profile holds mixed levels indefinitely.

docs/product/PRFAQ.md
mike-north: Done — level 1 is now 'environment variables holding plaintext secrets in your terminal or tool config', with .bashrc/mcp.json/.env named.

docs/product/PRFAQ.md
mike-north: Reframed around the relatable case: 'pass this key to this one command without leaving it lying around in a file, made routine' — and the difference from level 1 stated as no-longer-floating-on-disk for everything else to see.

docs/product/PRFAQ.md
mike-north: Rewrote with your framing: level 3 keeps level 2's only-the-chosen-process property but swaps the key for a lease, and the policy can be more granular than the target system's own permission model — the GitHub issues-only-with-this-label example is now in the text.

docs/product/PRFAQ.md
mike-north: You're right — 'cannot' was too flat. Rewrote: presence-bound keys (touch-required hardware policy attested at enrollment makes every valid signature proof of presence by construction) as the strong mechanism, a vault-signed assurance assertion for software-enforced backends with its trust root stated, and the whole thing positioned as roadmap rather than impossibility.

docs/product/PRFAQ.md
```diff
@@ -0,0 +1,160 @@
...
+verifying system's policy, not to vaultkeeper.
+
+### What about testing? Secrets tooling is notoriously untestable.
+
+vaultkeeper ships its own test doubles as products: an in-memory backend with real Ed25519
+signing and scriptable fault injection (throwing the same typed errors production throws), a
+presence simulator that is structurally unreachable from production code, and — in flight —
+backend-flavored doubles carrying real data shapes and choreography, held to the *same
+conformance corpus* as the real adapters. Manual testing reduces to a minutes-scale,
+tool-upgrade-triggered audit: "does the real adapter still match its double?"
```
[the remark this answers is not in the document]
mike-north: Reworded — the answer now distinguishes shipped (InMemoryBackend + TestVault), in review (signing, fault injection, presence simulator), and planned (flavored doubles), without claiming the richer doubles ship today.

docs/product/product-brief.md
```diff
@@ -0,0 +1,158 @@
...
+  signing leases, presence-at-session-establishment.
+- **The consuming toolmaker** — builds software (e.g. attestation tooling) on vaultkeeper
+  custody and needs to test its integration without hardware or accounts. Served by the shipped
+  test doubles and the paired-double conformance strategy.
+
+## 5. Core concepts
+
+- **Backend** — where secrets live. Pluggable: file (encrypted), macOS Keychain, 1Password,
+  YubiKey, DPAPI, libsecret. Backends self-report **capabilities**; policy fails closed when a
+  capability is absent rather than silently downgrading.
```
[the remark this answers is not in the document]
mike-north: Fixed — canonical name: secret-tool (Linux Secret Service via libsecret), in both docs.

docs/product/product-brief.md
```diff
@@ -0,0 +1,158 @@
...
+- **Capability token / lease** — an encrypted JWE granting scoped access: expiry, usage limit,
+  executable-identity binding at mint, revocation. The serialized, cross-process form is the
+  **lease**; "lease" is the house term for any serialized, expiring, revocable capability.
+- **Trust tier** — classification of the calling executable: `sigstore` (cryptographic
+  provenance) > `registry` (approved-hash manifest, TOFU) > `unverified`. Policy expresses a
+  minimum (`minTrust`), or-stronger.
+- **Presence** — a backend's ability to force a *fresh physical human action* for a specific
+  operation, right now (YubiKey touch, Touch ID, per-use biometric approval). Enforced
+  fail-closed; never simulated in production (the test simulator is structurally unreachable
+  there).
```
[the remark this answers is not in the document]
mike-north: Reworded as the enforced property plus a standing requirement on any test aid, without asserting a shipped simulator.

docs/product/product-brief.md
```diff
@@ -0,0 +1,158 @@
...
+- A lease's decisive property is off-box worthlessness — the leak class that actually happens
+  (log aggregators, telemetry, bug reports).
+- Non-invocability of a privileged identity requires OS-level isolation; custody cannot
+  substitute for it, and every custody recommendation states the precondition.
+- Fail-closed is the default posture: unknown backends report no capabilities; corrupt or
+  missing revocation state refuses lease validation; config naming an unavailable backend is an
+  error, never a silent fallback.
+- Test aids that could forge a security signal (the presence simulator) ship structurally
+  unreachable from production: separate dev-only package, no registry entry, opt-in-only
+  construction, loud refusal under production environments.
```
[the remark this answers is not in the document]
mike-north: Rephrased as a requirement ('must be structurally unreachable…'), noted as enforced by negative tests wherever such an aid exists.

docs/product/product-brief.md
```diff
@@ -0,0 +1,158 @@
...
+with bearer-safe redaction, encrypted key-state persistence, the environment-profile
+primitive, the lease-aware claims validator; the TypeScript SDK/CLI with the full backend
+suite; the WASM bridge with host-callback contracts; OIDC trusted publishing; conformance and
+packaging test suites.
+
+**In flight**: the `run` wrapper verb (stdio/signal-transparent); session signing leases with
+tamper-evident two-axis revocation; native backend parity (keychain via `security -i`, DPAPI,
+YubiKey ports); consumer test doubles (fault-injecting in-memory backend, presence simulator);
+the paired-double test strategy (stub framework, golden transcripts, flavored doubles, the
+manual-residue register).
```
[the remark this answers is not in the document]
mike-north: Reworded — 'signing and fault-injection capabilities being added to the existing in-memory backend', not implying the shipped one already has them.

docs/product/PRFAQ.md
```diff
@@ -0,0 +1,207 @@
...
+> Internal working document. The press release is aspirational-at-launch framing; the FAQ is
+> honest about what exists today versus what is in flight. Kept current by the maintainers.
+
+---
+
+## Press Release
+
+### vaultkeeper brings policy-enforced secret custody to every developer — using the credential store you already own
+
+**Developers can now get the secret-handling posture that previously required a business-tier
```
mike-north: Does this undersell the product a little bit? Even these business tier subscriptions to things like 1Password do not allow flexible policy to be defined. All they do is provide more powerful mechanisms. can access which secrets (e.g. 1p service accounts)

docs/product/PRFAQ.md
```diff
@@ -0,0 +1,207 @@
...
+on the table wherever demand appears, since a backend is a small adapter over a stable
+interface — and vaultkeeper supplies the governance layer those stores lack: capability tokens
+scoped by expiry, usage count, and executable identity; per-use human-presence requirements
+(a Touch ID tap or YubiKey touch *for this operation, right now*); delegated access patterns
+where the raw secret never enters the calling process at all; and full local revocation.
+
+Adoption is a gradient, not a rewrite, and the wrapper works on anything that reads
+environment variables. Wrap an MCP server so the PAT leaves the config file
+(`vaultkeeper run --profile github-mcp -- npx @github/mcp-server`); wrap an ordinary CLI for
+one policy-resolved invocation (`vaultkeeper run --profile deploy -- terraform apply`); or
```
mike-north, line 43: I think we need to cover "what is a profile?"

docs/product/PRFAQ.md
```diff
@@ -0,0 +1,207 @@
...
+keys go further still: the private key never leaves the backend at all; consumers request
+signatures, never key material.
+
+The same identity primitives serve human and automation signers alike. An agent fleet can hold
+signing identities whose keys are born inside vaultkeeper and cannot be extracted by the
+processes they govern — with human presence proven once at session establishment rather than
+demanded per signature.
+
+vaultkeeper is a single Rust core reachable from a native CLI and from TypeScript via
+WebAssembly, with a data-driven conformance suite holding every implementation to the same
```
mike-north: Rather than talk about TypeScript and WebAssembly, I think corporate environment friendliness is the value proposition here. WebAssembly and TypeScript are a means to an end. And that end is the ability to run in a corporate environment that has approved the use of Node.js (meaning it's compatible w/ things like google/santa). OSS means it's fully auditable

docs/product/PRFAQ.md
```diff
@@ -0,0 +1,207 @@
...
+### How is this different from dotenvx / `op run` / SOPS?
+
+Those tools solve *encrypted at rest, injected at runtime* — and solve it well. But they are
+not context-aware: they decrypt for whoever invokes them, so in practice you choose between
+security and friction (lock things down and pay for it on every invocation, or loosen them
+and hope). Because vaultkeeper's resolution step evaluates *policy* — who is asking, with
+what trust, for how long — it can offer both at once: strong defaults with low-friction
+day-to-day use. The concrete differences:
+
+1. **Policy at resolution** — trust tiers on the calling executable, TTLs, usage limits,
```
mike-north: > usage limits

what are we trying to articulate here? Just how many times a command may be invoked or are we trying to say more? When I think about usage limits, I think rate limits. but aren't we trying to say something bigger here something like custom rules (e.g. https://github.com/mike-north/celscript or similar)

docs/product/PRFAQ.md
```diff
@@ -0,0 +1,207 @@
...
+   per-use presence — not just decryption.
+2. **Storage-agnostic** — the same policy over Keychain, 1Password, YubiKey, DPAPI, the
+   Linux Secret Service, or a file, rather than one blessed store.
```
mike-north: Two value propositions here that we could call out more obviously. VaultKeeper works with the secret store you already use. And then the second is no lock-in. If you invest in scripting against VaultKeeper, you may deal with macOS keychain for some things and 1Password for other things and a YubiKey at work. And this gives you a unified facade

mike-north: I think this is particularly important for library and tool authors because it means you get to support whatever secret-keeping backend your user may have. And you don't have to deal with the complexity of manual testing flows and all of the intricacies of these different ways to store secrets. you can integrate with a secret keeping store just going through Vault Keeper and then your user gets to choose how they wish to protect these sensitive secrets and credentials.

docs/product/PRFAQ.md
```diff
@@ -0,0 +1,207 @@
...
+3. **The lease rung** — consumers can hold a capability instead of the secret, which nothing
```
mike-north: > capability

I believe this is the first place where we use the word capability, so we should have a parenthetical that helps a reader understand what we mean here.

docs/product/PRFAQ.md
```diff
@@ -0,0 +1,207 @@
...
+   Business tiers, locally and free.
+
+What those tools have that vaultkeeper deliberately does not: team sync and cloud
+distribution (see non-goals).
+
+### What is a VaultKeeper lease?
+
+An encrypted, serializable capability token that occupies the env-var slot *instead of* a
+secret. It carries expiry, a usage limit, and revocation state; redeeming it requires the
+local vault key (the encryption key vaultkeeper generates at setup and keeps in your
```
mike-north: > (the encryption key vaultkeeper generates at setup and keeps in your configured backend — only the vaultkeeper process on this machine can use it, enforced by the backend's own access control plus owner-only file permissions on the encrypted key state)

I think this is the right concept to communicate, but maybe it's going to read better if we frame it as almost a little mini story where in the beginning of this section we say when you set up Vault Keeper on a machine, it creates a key pair that only it has access to. We should probably have a name for this like "minting key" or something (feel free to brainstorm and choose a different name).

And then you can move on to articulating leases and the role this special key has in that process, why users should care about this, etc...

docs/product/PRFAQ.md
```diff
@@ -0,0 +1,207 @@
...
+configured backend — only the vaultkeeper process on this machine can use it, enforced by the
+backend's own access control plus owner-only file permissions on the encrypted key state), so
+a lease captured into a log aggregator or bug report is dead on arrival anywhere else. Signing
+leases are the same idea for signing keys: holding one lets a process *ask vaultkeeper to
+produce signatures* with a key it never sees — minted with human presence at launch, usable
+non-interactively until expiry, revocable individually or per-key.
+
+### How do I adopt this without rewriting my tools?
```
mike-north, line 130: I feel like we might want two sections here and one is oriented towards the tool user. And the real question here is how can I use VaultKeeper with my existing tools that have no knowledge of what VaultKeeper is? And that is roughly the section that you already have here. But then you might have a second section that's about library and tool authors who seek to offer pathways for secrets management and use. And this is more the embed case, using VaultKeeper as a library (potentially unbeknownst to the user of their tool/library)

docs/product/PRFAQ.md
```diff
@@ -0,0 +1,207 @@
...
+
+You don't adopt vaultkeeper wholesale — you upgrade **one secret at a time**, and each secret
```
mike-north:
```suggestion
You don't have to adopt vaultkeeper wholesale — you may upgrade **one secret at a time**, and each secret
```

docs/product/PRFAQ.md
```diff
@@ -0,0 +1,207 @@
...
+
+1. **Environment variables holding plaintext secrets in your terminal or tool config** (an
+   `export GITHUB_TOKEN=ghp_…` in `.bashrc`, a token pasted into `mcp.json` or a `.env`) —
+   the status quo being replaced. The secret sits on disk in plaintext, readable by anything,
+   forever.
+2. **Resolved at launch, never on disk.** You run the same tool through
+   `vaultkeeper run`, and the secret is fetched from your backend and handed *only to that
+   one process, for that one run*. If you have ever thought "I just need to pass this key to
+   this one command without leaving it lying around in a file" — this is that, made routine.
+   The tool is unmodified; the difference from level 1 is that the secret is no longer
```
mike-north:
```suggestion
   The tool is unmodified; the difference from using `.env` files with secrets in plaintext is that the secret is no longer
```
avoid referring to levels by number and just concisely state the pattern we're referring to

docs/product/PRFAQ.md
```diff
@@ -0,0 +1,207 @@
...
+   floating around on disk for every other process, backup, and sync client to see.
+3. **The process gets a lease, not the secret.** At level 2 the chosen process still holds
```
mike-north:
```suggestion
3. **The process gets a lease, not the secret.** When raw secrets (e.g. API keys) are made available to a process at launch, the chosen process still holds
```

docs/product/PRFAQ.md
```diff
@@ -0,0 +1,207 @@
...
+   direct access to. Those "right conditions" are yours to define: a time window, a usage
+   count, and — because vaultkeeper sits in the middle of every call — policy **more granular
+   than the target system itself can express**. GitHub's permission model cannot say "may
+   update issues, but only ones carrying this label"; a delegated fetch through vaultkeeper
+   can, because each invocation is evaluated against your policy before the key is ever used.
+   This level requires the consuming tool to speak vaultkeeper; today that means first-party
+   tooling, by choice (a language-agnostic local redemption endpoint is a named future epic,
+   gated on an installed base worth integrating against).
+
+### Does this work for AI-agent fleets?
```
mike-north, line 182: I might in this section also mention that in a world where agents are more and more involved in writing software, we also have a need to deterministically guard certain things that must be performed by humans. And VaultKeeper powers a solution for this as well through its sibling product, [`attest-it`](https://github.com/mike-north/attest-it)

docs/product/PRFAQ.md
```diff
@@ -0,0 +1,207 @@
...
+durable way; some agents (e.g. an adjudicator that signs policy-change proposals) need signing
+identities that the agents they govern cannot extract or invoke. vaultkeeper's answer: keys born
+inside a backend and never exported; signatures produced backend-side; session-scoped signing
+leases minted at harness launch with presence proven once; and honest documentation that
+*non-invocability* additionally requires OS-level isolation between the privileged harness and
+the processes it governs — custody cannot substitute for process isolation, and we say so.
+
+### Can a signature prove a human was present?
+
+Not in what ships today, but the path exists and is on the roadmap. Presence is enforced at
```
mike-north: > Not in what ships today

let's just add this to the roadmap rather than bake it into the PRFAQ as not shipping early on. This seems fairly important and we should just take on the work and write the PRFAQ from the perspective of it landing at launch.

docs/product/PRFAQ.md
```diff
@@ -0,0 +1,207 @@
...
+signing time by capable backends and is visible to the local caller; the resulting signature,
+however, does not yet carry that fact to a verifier. Two mechanisms can close the gap. The
+strongest is **presence-bound keys**: when a key lives in hardware whose signing policy
+requires a physical touch (a YubiKey with touch-required, a Secure-Enclave key gated on
```
mike-north: I might broaden this slightly so that it's clear. We strive to be able to integrate with a wide range of human-in-the-loop solutions. It's not restricted to requiring human touch. It could be a website with a captcha in front of an approval and the human has to click or they have to use a passkey or something like that. But the whole point is we're aiming for things that require human in the loop approval and an indication of human presence at the moment of approval (this is a solved problem in the esignature world, for example)

docs/product/PRFAQ.md
```diff
@@ -0,0 +1,207 @@
...
+
+### How do humans and automation coexist as signers?
+
+Identically. An identity is an Ed25519 keypair; nothing about being a human or an agent pins an
+identity to a role. What differs is **custody** — an attribute of the identity that can
+strengthen over time (file-backed today, presence-gated hardware tomorrow) without the identity
+becoming a different actor. Role semantics (who may author, who may approve) belong to the
+verifying system's policy, not to vaultkeeper.
+
+### What about testing? Secrets tooling is notoriously untestable.
```
mike-north: It's certainly not untestable, it's just annoying and difficult to test. And the idea is VaultKeeper has those annoying tests so you don't have to write them.

docs/product/product-brief.md
```diff
@@ -0,0 +1,168 @@
...
+  event.
+- **Delegated access** — `fetch`/`exec`/`sign` patterns where vaultkeeper performs the
+  operation and the raw secret never appears in the caller's memory; signing keys never leave
+  their backend at all.
+
+## 6. Architecture in one paragraph
+
+One semantic core in Rust (`vaultkeeper-core`), reached two ways: a native CLI, and TypeScript
+via a WASM bridge — with the TypeScript library's parallel logic being progressively retired as
+the core reaches parity (single-core consolidation). Backends are subprocess orchestration over
```
mike-north: > being progressively retired as the core reaches parity (single-core consolidation)

Frame this as an inevitability

docs/product/product-brief.md
```diff
@@ -0,0 +1,168 @@
...
+a host-platform abstraction wherever the underlying store has a CLI (keychain, secret-tool,
+DPAPI, YubiKey); 1Password remains host-provided because its per-process desktop-app grant is
+reachable only through a mechanism the WASM sandbox cannot load. A data-driven conformance
```
mike-north: >  1Password remains host-provided because its per-process desktop-app grant is reachable only through a mechanism the WASM sandbox cannot load.

No reason why this needs to be in a product brief

docs/product/product-brief.md
```diff
@@ -0,0 +1,168 @@
...
+
+## 7. How we make security claims
+
+The rule: **a claim appears in docs only if the code enforces it; a mechanism that only
+appears to deliver a property is treated as a defect.** Standing consequences:
+
+- Per-rung threat honesty. Rung 2 provides essentially zero confidentiality against same-UID
+  code execution; its real adversaries are exfiltration-at-rest and accidental disclosure. We
+  say this everywhere the value proposition is stated.
+- A lease's decisive property is off-box worthlessness — the leak class that actually happens
```
mike-north: > off-box worthlessness

is it really off-box or "outside of process" (could even be worthless _on the same machine_)?

docs/product/product-brief.md
```diff
@@ -0,0 +1,168 @@
...
+  (log aggregators, telemetry, bug reports).
+- Non-invocability of a privileged identity requires OS-level isolation; custody cannot
+  substitute for it, and every custody recommendation states the precondition.
```
mike-north: >  Non-invocability of a privileged identity requires OS-level isolation; custody cannot substitute for it, and every custody recommendation states the precondition.

This reads a bit like word salad. Explain more simply and directly

docs/product/product-brief.md
```diff
@@ -0,0 +1,168 @@
...
+- Fail-closed is the default posture: unknown backends report no capabilities; corrupt or
+  missing revocation state refuses lease validation; config naming an unavailable backend is an
+  error, never a silent fallback.
+- Test aids that could forge a security signal (e.g. a presence simulator) must be
+  structurally unreachable from production: separate dev-only package, no registry entry,
```
mike-north, line 118: We should even enforce that test aids return responses that have a very clear signal that they are test data. And then outside of the testing context, we could have explicit regret rejection of anything that looks like test data (e.g. all test leases begin with `test_` or something)

docs/product/product-brief.md
```diff
@@ -0,0 +1,168 @@
...
+  opt-in-only construction, loud refusal under production environments. This is a standing
+  requirement on any such aid, enforced by negative tests wherever one exists.
+
+## 8. Non-goals
+
+- **Team sync, cloud storage, cross-device distribution.** Subscription products earn their fee
+  there; competing would require becoming a cloud service and forfeiting "nothing leaves your
+  machine." Single-machine, single-developer parity is the scope.
+- **Locked-down corporate MCP environments** (proxy, cert auth, curated servers) — users there
+  cannot edit configs and do not have this problem.
```
mike-north: I'm not sure I would state this as a non-goal because our goal is to still make this tool usable in an MCP restricted environment. that we're specifically not trying to solve here is policy over tooling that does not not use a secret like an API key as the auth material.

docs/product/product-brief.md
```diff
@@ -0,0 +1,168 @@
...
+- **Matching 1Password's UX breadth.** Biometric ergonomics, browser integration, and sharing
```
mike-north: I think it would be more useful to broadly state that this is not intended to be a competing product to 1Password. In fact, this is intended to be complimentary and a powerful tool for those who use 1Password. If we feel that there is a need to state that matching 1Password's UX breadth is a non-goal, we have failed to position the product in such a way where it is obvious we are not trying to be a 1Password clone or competitor.

docs/product/product-brief.md
```diff
@@ -0,0 +1,168 @@
...
+  are not the competition; custody policy is.
+- **Modeling signature *semantics*** (authorship vs approval). What a signature means belongs to
+  the verifying system's per-gate policy; vaultkeeper makes identities distinguishable and
+  their keys non-transferable, and stops there.
```
mike-north: I might mention that it is possible to build these things on top of VaultKeeper, but VaultKeeper is aiming to be good at its job. And following the Unix tool philosophy, it's better for that added layer of responsibility around the meaning of a signature to belong to other tools that leverage Vaultkeeper.

docs/product/product-brief.md
```diff
@@ -0,0 +1,168 @@
...
+
+**Committed roadmap**: verifier-visible assurance — a signature provably presence-backed, via
+presence-bound hardware keys (signing policy attested at enrollment) and a vault-signed
+assurance assertion for software-enforced backends. Owner-decided 2026-07; presence here spans
+human-in-the-loop approval broadly (hardware touch, biometrics, passkey ceremonies,
+authenticated web approval), not only physical tokens.
+
+**Deliberately future**: the local redemption endpoint (opens rung 3 to any language);
+external key import with custody provenance modeled; issuance-side principal checks binding
+capabilities to the invoking context.
```
mike-north: This whole section feels like it is antithetical to the callout at the very top of this document, which aims to make this durable and a long-living artifact. If we're describing the state of the world and referring to specific minor releases here and what's in flight, we're going to have to update this with great regularity. Maybe every other day. And that is not the kind of life cycle I'm looking for from this kind of artifact.

docs/product/product-brief.md
```diff
@@ -0,0 +1,168 @@
...
+
+## 10. How we measure success
+
+- A new user gets a secret out of plaintext and behind policy in under five minutes, without
+  modifying the consuming tool.
+- The manual-testing residue stays a minutes-scale, tool-upgrade-triggered checklist — never a
+  suite that can rot unrun.
```
mike-north, line 179: I think I know what you're getting at here. And it can be better expressed as something like this (rephrase as you see fit): 

> Library and tool builders who have a need to integrate with the secrets backend, can achieve a high degree of release confidence without having to build and maintain their own manual test suites for things like 1password integration.

docs/product/product-brief.md
```diff
@@ -0,0 +1,168 @@
...
+presence-bound hardware keys (signing policy attested at enrollment) and a vault-signed
+assurance assertion for software-enforced backends. Owner-decided 2026-07; presence here spans
+human-in-the-loop approval broadly (hardware touch, biometrics, passkey ceremonies,
+authenticated web approval), not only physical tokens.
+
+**Deliberately future**: the local redemption endpoint (opens rung 3 to any language);
+external key import with custody provenance modeled; issuance-side principal checks binding
+capabilities to the invoking context.
+
+## 10. How we measure success
```
mike-north, line 169: I think we're missing a big one here, and that is VaultKeeper can provide vastly more secure secrets handling that can be brought to bear to use an extremely wide range of CLI and MCP tools.

docs/product/PRFAQ.md
```diff
@@ -0,0 +1,207 @@
+# vaultkeeper PRFAQ
+
+> Internal working document. The press release is aspirational-at-launch framing; the FAQ is
+> honest about what exists today versus what is in flight. Kept current by the maintainers.
```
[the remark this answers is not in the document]
mike-north: Reworded — the blockquote now frames both docs as the 1.0 definition written from launch perspective; 'Internal working document' is gone.

docs/product/PRFAQ.md
```diff
@@ -0,0 +1,207 @@
...
+1. **Environment variables holding plaintext secrets in your terminal or tool config** (an
+   `export GITHUB_TOKEN=ghp_…` in `.bashrc`, a token pasted into `mcp.json` or a `.env`) —
+   the status quo being replaced. The secret sits on disk in plaintext, readable by anything,
+   forever.
+2. **Resolved at launch, never on disk.** You run the same tool through
+   `vaultkeeper run`, and the secret is fetched from your backend and handed *only to that
+   one process, for that one run*. If you have ever thought "I just need to pass this key to
+   this one command without leaving it lying around in a file" — this is that, made routine.
+   The tool is unmodified; the difference from level 1 is that the secret is no longer
+   floating around on disk for every other process, backup, and sync client to see.
```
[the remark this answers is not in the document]
mike-north: Superseded by the 1.0 reframe — the doc now describes the 1.0 target throughout, with the 'Is everything here real today?' anchor pointing at the tracker for current status; no exec/run hedging remains.

docs/product/PRFAQ.md
mike-north: Agreed — headline now reads 'get — and go beyond —' with per-invocation policy called out as something no subscription tier offers at any price; body notes paid tiers strengthen who-may-access mechanisms without offering conditions-of-use policy.

docs/product/PRFAQ.md
mike-north: Added a 'What is a profile?' FAQ — named per-project file, env vars → secret source + materialization + per-entry policy; names and policy only, never values; committable so weakening becomes a reviewable diff.

docs/product/PRFAQ.md
mike-north: Rewrote the paragraph around corporate-environment friendliness: runs where Node.js is already approved (Santa-compatible), native binary where permitted, fully auditable OSS — TS/WASM demoted to the means.

docs/product/PRFAQ.md
mike-north: Reworded to 'policy at resolution, per invocation' — time windows, use counts, presence — with custom rule expressions named as the direction (policy as expressive as your rules, not as coarse as a permission checkbox).

docs/product/PRFAQ.md
mike-north: Both called out: item 2 is now 'works with the store you already use, with no lock-in' — one unified facade, mixing Keychain/1Password/YubiKey freely, changing stores never changes tooling.

docs/product/PRFAQ.md
mike-north: Added the parenthetical at first use: a capability is a token that permits specific operations to be performed on the holder's behalf — not the secret itself.

docs/product/PRFAQ.md
mike-north: Rebuilt as the mini-story: setup creates the minting key (name adopted — it pairs with leases being minted/redeemed); then leases, the key's role, and why the reader should care.

docs/product/PRFAQ.md
mike-north: Split as suggested: the tool-user question ('How do I adopt this without rewriting my tools?') and a new library/tool-author section covering the embed case — integrate once, support whatever store the user has, potentially unbeknownst to them.

docs/product/PRFAQ.md
mike-north: Applied as suggested.

docs/product/PRFAQ.md
mike-north: Applied — level-number references removed in favor of naming the pattern (.env files with plaintext secrets).

docs/product/PRFAQ.md
mike-north: Applied as suggested.

docs/product/PRFAQ.md
mike-north: Added — the agent-fleet answer now closes with the deterministically-human need and links attest-it as the sibling that powers it, with vaultkeeper holding the signing keys.

docs/product/PRFAQ.md
mike-north: Done, and generalized: the whole document is now explicitly the 1.0 PRFAQ, written from launch perspective without hedging. The assurance answer is affirmative ('Yes — at two strengths'), and #320 tracks the work (both mechanisms, trust roots, attest-it wire-format coordination).

docs/product/PRFAQ.md
mike-north: Broadened — presence now explicitly spans human-in-the-loop approval generally: passkey ceremonies, authenticated web approval behind a CAPTCHA, the e-signature-world pattern; the point is trustworthy indication of human involvement at the moment of approval, not touch specifically.

docs/product/PRFAQ.md
mike-north: Adopted nearly verbatim: not untestable, annoying and difficult — vaultkeeper owns the annoying tests so consumers don't write them.

docs/product/product-brief.md
mike-north: Done — 'will be retired; single-core consolidation is the settled direction, not an open question.'

docs/product/product-brief.md
mike-north: Cut — implementation detail at the wrong altitude for this document. Backends are 'small adapters over a host-platform abstraction' and the wasm/dlopen specifics stay in the engineering record.

docs/product/product-brief.md
mike-north: Good push. Under the 1.0 framing this is now stated as the stronger property: worthless outside its intended holder — off-box it cannot be decrypted, and same-machine redemption verifies the presenter, so a lease lifted by another process is refused. (Pre-1.0 reality: only the off-box half is enforced today; the presenter check is the principal-check work, noted on #282 as 1.0 scope.)

docs/product/product-brief.md
mike-north: Rewritten in plain speech: holding the only copy of a key doesn't help if any process on the machine can simply ask vaultkeeper to use it — so a privileged identity's governed processes must run under a different OS user, and every custody recommendation says so up front.

docs/product/product-brief.md
mike-north: Adopted as a new standing rule alongside the unreachability guards: test artifacts are self-announcing (marked leases etc.) and non-test code paths explicitly refuse marked artifacts. Unreachability protects against test code leaking; this protects against test data leaking. Work item being filed, cross-referenced to the in-flight simulator PR so the claim shape isn't foreclosed.

docs/product/product-brief.md
mike-north: Corrected — the non-goal is now 'governing auth that isn't secret-material' (client certs, platform-attested identity): no key to custody, no policy to enforce. Wherever an API-key-shaped secret exists — including locked-down corporate environments — vaultkeeper aims to be usable.

docs/product/product-brief.md
mike-north: Reframed affirmatively: 'Competing with 1Password — vaultkeeper is complementary to it' — one of vaultkeeper's best backends, and a power tool FOR 1Password users. The UX-breadth disclaimer is gone; you're right that needing it would mean the positioning already failed.

docs/product/product-brief.md
mike-north: Adopted with the Unix framing: such things can absolutely be built on top — and are, attest-it first among them; that layer of meaning belongs to tools that leverage vaultkeeper, and vaultkeeper aims to be excellent at its own job and stops there.

docs/product/product-brief.md
mike-north: Fully agreed — this was antithetical to the document's premise. The release-inventory section is gone from both docs, replaced by 'What 1.0 means' (scope pillars a reader can hold us to) in the brief and the 'Is everything here real today?' anchor in the PRFAQ; the tracker owns current status.

docs/product/product-brief.md
mike-north: Adopted nearly verbatim as a success measure: library and tool builders achieve a high degree of release confidence without building or maintaining their own manual test suites for things like 1Password integration.

docs/product/product-brief.md
mike-north: Added as the headline measure: dramatically more secure secrets handling brought to bear across an extremely wide range of CLI and MCP tools — breadth of tools running under vaultkeeper leads the list.

docs/product/PRFAQ.md
```diff
@@ -0,0 +1,259 @@
...
+Service (`secret-tool`), or an encrypted file store, with further stores (Dashlane and others)
+on the table wherever demand appears, since a backend is a small adapter over a stable
+interface — and vaultkeeper supplies the governance layer those stores lack: capability tokens
+scoped by expiry, usage count, and executable identity; per-use human-presence requirements
+(a Touch ID tap or YubiKey touch *for this operation, right now*); delegated access patterns
+where the raw secret never enters the calling process at all; and full local revocation.
+
+Adoption is a gradient, not a rewrite, and the wrapper works on anything that reads
+environment variables. Wrap an MCP server so the PAT leaves the config file
+(`vaultkeeper run --profile github-mcp -- npx @github/mcp-server`); wrap an ordinary CLI for
```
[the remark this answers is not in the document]
mike-north: Declining this one deliberately: the pnpm-only convention governs contributors working in this repository, but this example depicts an end user's mcp.json — and npx is the ecosystem-standard invocation in MCP server documentation everywhere. A product example should read as what users actually write; pnpm dlx here would make it less recognizable to the reader it serves.

docs/product/product-brief.md
```diff
@@ -0,0 +1,161 @@
...
+
+| Rung | Env var contains | Works with | What it buys |
+|---|---|---|---|
+| 1 | The plaintext secret, from a file | Everything | Nothing. The status quo being replaced. |
+| 2 | The real secret, resolved at launch | **Any unmodified tool** | No plaintext at rest; exposure bounded to a process lifetime; a policy decision (trust tier, TTL, presence) at every resolution; centralized rotation; kills the copy-paste-into-files habit. |
+| 3 | A **VaultKeeper lease** — an encrypted, expiring, revocable capability | vaultkeeper-aware consumers | The secret never enters the consumer's process; the lease is worthless off-box and after expiry; usage-limited; revocable per-lease and per-key. |
+
+The rung is chosen **per entry** in a profile, and mixing is the permanent steady state, not a
+migration phase. Rung 2's delivery mechanism is the `vaultkeeper run` wrapper — most visibly as
+the `command` in an MCP server config. Rung 3 is first-party-only today *by choice*; the
```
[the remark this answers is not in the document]
mike-north: (Adding the reply this thread was resolved without.) Addressed in two steps: first with an exec-today/run-in-flight parenthetical, then superseded by the 1.0 reframe — the brief now describes the 1.0 target throughout and no longer claims to document the shipped CLI surface, which dissolves the inconsistency this thread flagged.

docs/product/PRFAQ.md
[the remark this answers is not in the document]
mike-north: Adjudicated as not an issue: the npx here is the end user's own MCP-server launch command being wrapped by vaultkeeper run — npx is the MCP ecosystem's standard invocation, and user-facing docs should mirror real usage. The repo's pnpm-only convention governs development within this repository, not what users type.

Issue #299, opened 2026-07-22
Implement session mint for signing leases with mandatory presence-at-mint


mike-north:
Refs #276

Implement `session mint`: mint a **session signing lease** — a serializable, expiring, revocable, presence-minted authorization to use a signing key — and enforce presence at mint time. This closes the asymmetry where the secret path has a wire form, `exp`, `use`, and a caller check while the in-memory signing handle (`authorizeSigningKey()`) has none.

## The lease shape

Same JWE envelope as `setup()` — encrypted under the vault key via `KeyManager`, so rotation and `revokeKey()` already apply. Distinct claims payload, with **no `val`**:

```
{ jti, iat, exp, sub: <key name>, exe, tid, use, kty: "signing-key",
  kid: <public key id>, ref: <backendRef>,
  kgen: <key generation at mint>,
  pres: { op: "sign", at: <unix>, method: "backend-touch" | "host-approval", backend: "<type>" } }
```

- `exp` is Unix **seconds**, minted from the entry's `ttlSeconds`. Default TTL **28800 (8h)**, hard cap **86400 (24h)**.
- `kgen` is stamped from the **current key generation** at mint (from the revocation-store key-generation map) so the per-key revocation axis works.
- The claims already validate through the `kty`-discriminated `validateClaims` (merged): `val` is not required for a signing lease.

## Presence-at-mint

`pres` is populated **only** when the mint path actually proved presence, via one of two existing mechanisms:

- `method: "backend-touch"` — `enforce_presence_requirement(backend, PresenceOperation::Sign, Some(true))` succeeded; the backend advertises `BackendCapabilities::enforces(Sign)` and forced a fresh physical action.
- `method: "host-approval"` — `HostPlatform::prompt_approval` returned `true`. Its default is `Ok(false)` (fails closed) — correct for a host with no interactive approval path.

**If neither is available, `session mint` fails** with a typed `NotCapable` error. There is **no mint-without-presence mode** and no silent downgrade. A backend reporting no `Sign` presence must produce `NotCapable`, never a lease with an empty or fabricated `pres`.

## Non-interactive / unattended-restart rule

[...]

PR #328, opened 2026-07-22
Add native CLI session mint for signing-key leases with presence-at-mint

mike-north:
## Summary

Implements `vaultkeeper session mint --profile <NAME> --entry <VAR>` in the native Rust CLI: mints a session signing-key lease — a serializable, expiring, revocable, presence-minted authorization to use a signing key — closing the asymmetry where the secret path already has an `exp`/`use`/caller check and the signing path did not.

- `VaultKeeper::mint_signing_lease` (new, `crates/vaultkeeper-core/src/vault.rs`) mints the same JWE envelope `setup()` uses (encrypted under the vault key via `KeyManager`, so rotation/`revokeKey()` already apply), with a distinct claims payload carrying no `val`: `jti, iat, exp, sub, exe, tid, use, kty:"signing-key", kid, ref, kgen, pres`.
- Presence-at-mint is proven via exactly one of two mechanisms, never both, never neither with a fabricated `pres`:
  - `"backend-touch"` — `enforce_presence_requirement(backend, PresenceOperation::Sign, Some(true))` succeeds.
  - `"host-approval"` — `HostPlatform::prompt_approval` returns `Ok(true)`, tried only after backend-touch fails and only when the invocation is interactive.
  - Neither available → `VaultError::NotCapable`.
- **Non-interactive fail-closed rule**: when presence is required and the invocation is non-interactive (native CLI: `stderr` is not a terminal), `prompt_approval` is **never called at all** — a structural guarantee against an unbounded hang on a future interactive host, not just a race against its default timing. The exact documented error message is reproduced verbatim.
- `kgen` is stamped from the signing key's current generation, read fresh from the persisted revocation store (`crate::keys::load_revocation_for_validation`) on every mint — never cached, never defaulted.
- `exp` derives from the entry's `ttlSeconds` (default 28800s / 8h), defensively capped at the mint layer to `SIGNING_LEASE_MAX_TTL_SECONDS` (86400s / 24h) in addition to the profile loader's existing hard rejection of an over-cap value.
- The native CLI's `session mint` resolves the profile/entry (reusing the existing profile loader), rejects a secret-backed or non-`lease` entry, computes `interactive` from `stderr`'s tty-ness, and threads it all into `mint_signing_lease` — printing the JWE to stdout on success.

Rust is touched, so the committed WASM artifact was rebuilt (`wasm-pack build --target nodejs`, pinned toolchain versions from `crates/vaultkeeper-wasm/wasm-toolchain.env`) and the fingerprint check (`scripts/wasm-export-fingerprint.mjs`) confirms the export/import surface is unchanged (47 exports / 54 imports, matching) — only the `.wasm` binary size changed, well within budget.

1. **(unit)** backend reporting no `Sign` presence → `NotCapable`, no fabricated `pres` — `mint_signing_lease_tests::backend_with_no_sign_presence_fails_with_not_capable` and `backend_that_enforces_only_other_operations_fails_with_not_capable` (`crates/vaultkeeper-core/src/vault.rs`).
2. **(unit)** `pres.method` reflects the mechanism actually used — `pres_method_is_backend_touch_when_the_backend_enforces_sign` and `pres_method_is_host_approval_when_the_backend_is_incapable_but_host_approves`.
3. **(unit)** `exp` derived from `ttlSeconds` (default 28800; a requested TTL >86400 is capped, not rejected, at this layer) and `kgen` equals the current key generation — `exp_defaults_to_the_eight_hour_ttl`, `a_requested_ttl_above_the_hard_cap_is_capped_not_rejected`, `kgen_equals_the_current_key_generation_at_mint`.
4. **(integration/UAT)** a non-interactive invocation of a presence-requiring profile fails closed with the documented message, returns promptly, never blocks — `session_mint::presence_requiring_entry_fails_closed_when_non_interactive` (`crates/vaultkeeper-cli/tests/cli_integration.rs`; spawns the real compiled binary via `assert_cmd` with a bounded timeout and asserts elapsed time, exit code, and exact stderr text).
5. **(integration)** a non-presence profile resolves unattended with no prompt — `session_mint::non_presence_entry_mints_unattended_with_no_prompt`.

Additional coverage: `no_presence_required_mints_with_no_pres_and_no_prompt`, `non_interactive_invocation_fails_closed_without_ever_calling_prompt_approval` (asserts the prompt call counter stays at 0), `minted_lease_carries_no_val_and_is_kty_signing_key`, and CLI-surface negatives (`mint_fails_when_the_profile_does_not_exist`, `mint_fails_when_the_entry_does_not_exist`, `mint_fails_for_a_secret_backed_entry`, `mint_exits_2_when_required_flags_are_missing`, `session_help_documents_mint_subcommand`).

## Test plan

- [x] `cargo test --workspace` — all green (324 vaultkeeper-core tests including 10 new `mint_signing_lease_tests`, 85 vaultkeeper-cli tests including 10 new `session_mint` tests, no failures)
- [x] `cargo clippy --all-targets -- -D warnings` — clean
- [x] `cargo fmt --check` — clean
- [x] `pnpm check` — clean (pre-existing `security/detect-object-injection` warnings only, unrelated to this change)
- [x] `pnpm test` — 265 passed, 19 skipped (pre-existing), 0 failed
- [x] WASM rebuilt and export/import fingerprint verified unchanged against the previously committed artifact

Refs #299


mike-north:
Review complete; fix batch pushed as 2974312, all 11 threads resolved (dispositions here since the pending review blocks inline replies). Fixed: explicit decline now surfaces PresenceDeclined (variant constructed for the first time, prompt-invocation asserted); only genuine NotCapable falls through the backend-touch probe — any other failure propagates without prompting, in both interactive and non-interactive modes; the anti-rollback watermark is consulted and updated at mint (no more hardcoded zero); the mintable policy is enforced inside VaultKeeper::mint_signing_lease itself with core-level cheat tests; stale doc param fixed; LeasePresence construction deduped; session help lists mint; PROFILE is positional matching the sibling commands; the no-entry error hints profile show; the TTL constants are pinned to the spec literals (28800/86400 with issue #299 references); and the mint e2e now decrypts the emitted JWE and asserts kty/kid/kgen present with no val. One thread adjudicated as a false positive: the \u{20} escapes in the wrapped message string are deliberate — Rust's backslash-newline continuation strips leading whitespace (verified with a standalone rustc compile), so the escapes re-add exactly the intended spacing.

Issue #312, opened 2026-07-22
Add conformance fidelity gate for the test double plus core in_memory signing parity


mike-north:
Refs #306

Hold the `@vaultkeeper/test-helpers` `InMemoryBackend` double to the same contract corpus as the real backends, in vaultkeeper's own CI, and mirror the signing + presence capabilities on the core Rust `in_memory` backend for parity. This is what makes the double "authoritative by construction, versions in lockstep" a mechanically-enforced property rather than a promise: when real backend behavior moves, the double's tests move with it.

### 1. Conformance fidelity gate against the double

- Run the applicable data-driven conformance cases against the TS `InMemoryBackend` double in vaultkeeper's own test suite, following the existing JS conformance-runner pattern in `packages/cli-tests/test/conformance/` (cases exported from `crates/vaultkeeper-conformance`). The double is held to the same contract corpus the real backends are, so a drift between the double and the real contract fails CI.
- Scope the run to the cases actually applicable to an in-memory backend (store / retrieve / sign / capability behavior); cases that require a specific real-backend substrate are out of scope for the double.

### 2. Core `in_memory` signing + capability parity

- Mirror signing and presence capabilities on the core Rust backend `crates/vaultkeeper-core/src/backend/in_memory.rs`, implementing the existing `SigningBackend` trait so the Rust `InMemoryBackend` matches the TS double. Use the file-backend's in-crate signing implementation as the model. This keeps the two-language backends at parity (core-first hygiene) and lets the conformance corpus exercise the same behavior on both sides.

### 3. Manual-residue register (committed, release-referenced doc)

As the cross-cutting trust anchor for the paired-double strategy, this item also commits the **manual-residue register** — the complete, auditable human residue that remains once every backend has a paired double running the shared corpus. It belongs here rather than in any single backend's item because it spans all five backends.

When all tiers land, **each backend's residue is one line: run the shared contract corpus against the real adapter** (the double's exact CI corpus, re-pointed at the real tool/account/hardware) — plus, only where a biometric/hardware ceremony gates it, a physical tap. The register states cadence and cost so it visibly cannot rot the way six-month manual suites did:

| Backend | Manual residue = run shared corpus vs. real adapter | Physical ceremony | Cadence · cost |
|---|---|---|---|
| keychain | corpus vs. real `security` on macOS | Touch-ID-gated read tap | per `security`/OS bump · ~2 min |
[...]

Issue #310, opened 2026-07-22
Extend @vaultkeeper/test-helpers: InMemoryBackend signing + presence capability surface, fault injection, TestVault signing ceremony


mike-north:
Refs #306

Grow the `@vaultkeeper/test-helpers` package (`packages/test-helpers/`) so a consumer can wire a hardware-free, CI-runnable test double into its own automated tests in place of a real 1Password / Keychain / hardware-backed store — covering the store → retrieve → **sign** path and driving failure modes deterministically. This package is already a separate, dev-only npm package (never a runtime dependency, absent from the main bundle by construction), which is the correct delivery surface.

## Motivation

The consuming project verifies its backend integration today through manual, human-run tests requiring physical hardware (a YubiKey with HMAC challenge-response, an authenticated `op` session, macOS Keychain). Those gates have not run in roughly six months and were reclassified as known flakes — so automated verification of the integration seam is effectively not happening. A double shipped by the owner of the behavior is authoritative by construction and versions in lockstep with the thing it doubles, unlike a consumer-authored fake that rots silently when real behavior moves. The goal is to catch "the consumer calls vaultkeeper wrong," not to re-test vaultkeeper internals.

### 1. `InMemoryBackend` implements `SigningBackend` and `PresenceCapableBackend`

- Implement the `SigningBackend` interface from `packages/vaultkeeper/src/backend/types.ts` (`generate` / `getPublicKey` / `sign`; the private key never leaves the backend). Use **real in-memory Ed25519 keys producing real signatures** that verify against the exported public key — fidelity sufficient to catch wiring errors, zero hardware. Model the implementation on the existing TS signing implementation in the file backend (`packages/vaultkeeper/src/backend/file-backend.ts`).
- Implement `PresenceCapableBackend` (same `types.ts`), **reporting no-presence by default** — aligned with the existing `BackendCapabilities` vocabulary, introducing no parallel concept.

### 2. Deterministic, scriptable per-operation fault injection

- A fault-injection API on the double: per-operation, scriptable, supporting both **one-shot** (fires once, then clears) and **persistent** (fires until removed) modes.
- Injected faults must throw the **real typed error classes** from `packages/vaultkeeper/src/errors.ts` — not stringly-typed lookalikes — so a consumer's `catch` blocks see in tests exactly what they see in production. Cover the four named modes: **backend unavailable, key absent, permission denied, session expired** (map each to its real typed class in the hierarchy).

[...]

Issue #273, opened 2026-07-21
Rust CLI: wire BackendRegistry into main.rs so commands (incl. backend capabilities) enumerate registered backends


mike-north:
## Context

PR #266 added `backend capabilities` to the Rust CLI, but it reports exactly one row (the hardcoded active `FileBackend`) while the TS CLI enumerates all six registered backend types via `BackendRegistry`. The divergence is documented and deliberate in #266 — the root cause is that no config-driven backend registry is wired into `crates/vaultkeeper-cli/src/main.rs` at all (`store`/`delete`/`doctor` also hardcode `FileBackend`).

This breaks the repo's identical-command-surface convention for `--json` consumers: a script checking whether e.g. `yubikey` supports presence-per-use silently finds no matching row against the Rust CLI instead of an answer.

## What

Wire `vaultkeeper_core::backend::BackendRegistry` into the Rust CLI's backend resolution (part of epic #234's consolidation), then widen `backend capabilities` to enumerate registered backend types like the TS CLI (six rows, same names/order), updating the conformance exact-output pins and the `--help` text that currently says "active (configured) backend".

## Verify

- Both CLIs' `backend capabilities --json` parse to the same set of rows for the backends available on the platform.
- Conformance cases updated and regenerated (`export-conformance`); JS runner green.

mike-north:
Additional acceptance criterion (from the backend-consolidation design, Refs #288): when `config.json` names a backend type the binary cannot provide, the CLI must fail closed with a typed `BackendUnavailable`-class error naming the requested backend — never silently fall back to the file backend. A silent fallback is a confidentiality downgrade disguised as success. Negative test required: config requesting an unregistered backend type → typed error, non-zero exit, no file-backend I/O performed.

PR #329, opened 2026-07-22
Conformance fidelity gate for InMemoryBackend + core in_memory signing/capability parity

mike-north:
## Summary

Adds a conformance fidelity gate that holds the `@vaultkeeper/test-helpers` `InMemoryBackend`
double to the same contract corpus the real backends are held to, mirrors the corresponding
signing + presence capability surface onto the Rust core `InMemoryBackend`, and commits the
manual-residue register documenting the human residue that remains once every backend has a
paired double running that shared corpus.

The existing `packages/cli-tests/test/conformance/` corpus is CLI-argv-only and always exercises
the hardcoded `file` backend (see #273/#297) — there is no CLI invocation that selects a specific
backend instance, so it cannot exercise the `InMemoryBackend` double directly. Rather than
generalizing that corpus (explicitly out of scope per the issue), this adds a second, narrower
data-driven corpus in `crates/vaultkeeper-conformance/src/backend_cases.rs`, scoped to exactly the
store/retrieve/delete/exists/list/sign/capability behavior an in-memory backend can support, and
runs it in-process against both languages' `InMemoryBackend`.

## What changed

- `crates/vaultkeeper-conformance/src/backend_cases.rs` (new): 11 data-driven
  `BackendConformanceCase`s (store/retrieve round-trip, not-found errors, delete, exists, list,
  signing key generate/sign/verify, duplicate-key rejection, missing-key errors, presence
  capability), exported to JSON via a new `export-backend-conformance` bin.
- `crates/vaultkeeper-core/tests/backend_conformance.rs` (new): runs the same corpus against the
  Rust core `InMemoryBackend`.
- `packages/test-helpers/test/conformance/backend-cases.{json,test.ts}` (new): runs the same
  corpus (committed export of the Rust corpus) against the TS `InMemoryBackend` double, following
  the existing JS conformance-runner pattern (`packages/cli-tests/test/conformance/run-conformance.test.ts`).
- `crates/vaultkeeper-core/src/backend/in_memory.rs`: implements `SigningBackend` (real in-memory
  Ed25519 keys via `ed25519-dalek`, mirroring `FileBackend`'s in-crate signing implementation) and
  `PresenceCapableBackend` (honest `presence_per_use: false`), matching the TS double's existing
  surface (#310). Also fixes `delete()` on a missing id, which previously silently no-op'd instead
  of returning `SecretNotFound` — a real divergence from `FileBackend`'s contract and the TS
  double's `delete()`, caught while building the new cross-language corpus.
- `crates/vaultkeeper-core/tests/presence_capability.rs`: updates two tests that used
  `InMemoryBackend` as their "backend without `PresenceCapableBackend`" example (no longer true
  now that it implements the trait); adds a `PlainBackend` fixture for that purpose and a new test
  asserting `InMemoryBackend` now correctly reports capability via the trait.
- `docs/manual-tests/manual-residue-register.md` (new): the five-backend register from the issue
  body, the tool-upgrade-triggered (not calendar-triggered) framing, the physical-ceremony column,
  and why each ceremony can't be automated. Referenced from the release workflow
  (`.github/workflows/release.yml` header comment) and cross-linked from
  `docs/manual-tests/presence-per-use.md` and `packages/vaultkeeper/README.md`.
- `packages/vaultkeeper-wasm/wasm/vaultkeeper_wasm_bg.wasm`: rebuilt (Rust core changed); export/import
  fingerprint unchanged (`node scripts/wasm-export-fingerprint.mjs` — 47 exports/54 imports both
  sides), well within the gzip/uncompressed size budgets.

1. **(conformance)** applicable cases run against the TS `InMemoryBackend` double and pass — `packages/test-helpers/test/conformance/backend-cases.test.ts` (`InMemoryBackend conformance fidelity gate` describe block, 11 cases via `it.each`).
2. **(conformance)** a deliberate divergence causes the gate to fail, proving it isn't vacuous — `backend-cases.test.ts`'s `InMemoryBackend conformance fidelity gate — drift detection (AC2)` block: two tests that mutate a case in-memory (wrong expected secret value; incorrect not-found assertion) and assert the runner rejects. The mutation only exists inside the test — no broken case ships in the corpus.
3. **(unit, Rust)** core `in_memory` implements `SigningBackend`: generates a keypair, signs, and the signature verifies against the public key — `crates/vaultkeeper-core/src/backend/in_memory.rs` `tests` module (`sign_with_key_produces_a_signature_verifiable_against_the_public_key`, plus duplicate/missing-key/tamper/coexistence/clear regression tests).
4. **(conformance)** the same cases pass against the core Rust `in_memory` backend, confirming cross-language parity — `crates/vaultkeeper-core/tests/backend_conformance.rs` (`all_backend_conformance_cases_pass_against_core_in_memory_backend`).
5. **(UAT/doc)** the manual-residue register is committed and referenced by the release process, with the five rows, the tool-upgrade-triggered framing, the ceremony column, and cadence·cost — `docs/manual-tests/manual-residue-register.md`, referenced from `.github/workflows/release.yml`'s header comment.

## Test gates

- `cargo test --workspace`: 538 tests passed, 0 failed across all 19 test binaries (unit tests, `crates/vaultkeeper-core/tests/*`, the new `backend_conformance` binary, the `vaultkeeper-conformance` crate's own unit tests, etc.).
- `cargo fmt --check`: clean.
- `pnpm check` (typecheck + lint + api-report + api-docs across all TS packages): clean.
- `pnpm test`: 1012 + 7 + 102 + 355 + 265 = 1741 tests passed, 0 failed (24 skipped, environment-gated) across `vaultkeeper`, `@vaultkeeper/wasm` (Node's `node:test`, 131/131), `@vaultkeeper/test-helpers` (102/102, including the new 13-test conformance file), `@vaultkeeper/cli`, `@vaultkeeper/cli-tests`.

## Deferred follow-ups

- None identified in scope. The pre-existing CLI-argv-only conformance corpus (#273/#297) remains
  as-is per the issue's explicit "do not block on generalizing the corpus."

Refs #312

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,6 +31,23 @@ jobs:
       - name: Test
         run: cargo test
 
+      - name: Verify committed backend conformance corpus matches a fresh export (drift check)
+        # The Rust corpus in crates/vaultkeeper-conformance/src/backend_cases.rs
+        # is the single source of truth; the JS conformance runner in
+        # packages/test-helpers/test/conformance/backend-cases.test.ts loads a
+        # committed export of it rather than regenerating it at test time (so
+        # the TS package has no Cargo dependency at test time). If someone
+        # edits the corpus without re-running the exporter, the committed JSON
+        # silently goes stale and the TS suite stops exercising the real
+        # cases. Mirrors the check:api-report regeneration-diff pattern.
+        run: |
+          cargo run -q -p vaultkeeper-conformance --bin export-backend-conformance \
+            > /tmp/backend-cases-fresh.json
+          if ! diff -u packages/test-helpers/test/conformance/backend-cases.json /tmp/backend-cases-fresh.json; then
+            echo "::error::packages/test-helpers/test/conformance/backend-cases.json is stale. Regenerate it: cargo run -p vaultkeeper-conformance --bin export-backend-conformance > packages/test-helpers/test/conformance/backend-cases.json"
+            exit 1
+          fi
+
       - name: Build release binary
         run: cargo build --release
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,3 +1,9 @@
+# Before a release that bumps a backend's underlying tool/OS/SDK dependency
+# (`security`, the 1Password desktop app/`op`, `ykman`, `secret-tool`/
+# `gnome-keyring`, `powershell`/Windows), run the shared conformance corpus
+# against that backend's real adapter and record the result in the release
+# notes — see docs/manual-tests/manual-residue-register.md for the full
+# per-backend register (issue #312).
 name: Release
 
 on:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -131,7 +131,7 @@ jobs:
 
       - name: Manual-residue register reminder
         run: |
-          echo "::notice::Before publishing a release that bumps a backend's underlying tool/OS/SDK dependency (security, the 1Password desktop app/op, ykman, secret-tool/gnome-keyring, powershell/Windows), run the shared conformance corpus against that backend's real adapter and record the result in docs/manual-tests/manual-residue-register.md (issue #312)."
+          echo "::notice::Before publishing a release that bumps a backend's underlying tool/OS/SDK dependency (security, the 1Password desktop app/op, ykman, secret-tool/gnome-keyring, powershell/Windows), run the shared conformance corpus against that backend's real adapter and record the outcome in the release notes, per docs/manual-tests/manual-residue-register.md (issue #312)."
           echo "See: docs/manual-tests/manual-residue-register.md"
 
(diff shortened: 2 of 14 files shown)


mike-north:
Review complete; fix batch pushed as f5b2f89 and all 9 threads resolved (dispositions, since inline replies are blocked by the pending review): the gate was empirically proven non-vacuous by inducing a lenient-delete drift in the TS double and watching the fidelity case fail; the Ed25519 seed is now zeroized after key construction (three threads, one fix); the TS runner uses err.name, an exhaustiveness-guarded switch, and runtime shape validation of the parsed corpus; the Rust runner gained its own mutate-a-case drift test; the corpus gained a list-omits-after-delete case through a new negative list assertion with the export regenerated byte-identical; the release job now echoes the manual-residue register pointer into every run's log; and a CI step diffs a fresh corpus export against the committed JSON, closing the previously-untracked sync gap between the two runners' case lists.

Issue #300, opened 2026-07-22
Redeem session signing leases through the core handle table with single-source accounting


mike-north:
Refs #276
Refs #282

Implement redemption of a session signing lease into a usable core signing handle, with **all lifetime and use accounting done exactly once, in core** — the single accounting authority. This is the read side of the lease: a lease presented back to the issuer is validated across every axis and, if good, produces a bounded handle that can sign up to its use limit until it expires or is revoked.

## What to build

Redeem lease (JWE) → core signing handle. `exp` and `use` are enforced by the **core handle table** (merged), **not** a parallel mechanism. There must be **no second accounting path**: the handle table is the single accounting authority for both expiry and use count. Redemption validates the lease and, on success, installs a handle whose lifetime and remaining uses are tracked in that one place.

## Gates this item must respect (not lift)

- **Finite expiry is mandatory.** This item does **not** reopen never-expiring handles — every redeemed lease handle carries a finite `exp` and the handle table enforces it. Do not add a path that produces an unbounded-lifetime handle.
- **The principal / caller check remains in force** (Refs #282). Redemption must respect that gate, not bypass or weaken it.

## Validation axes (each a distinct typed error, no panics)

On redemption, reject with the correct typed error:

- **Expired** — lease past its `exp`.
- **Exhausted** — lease whose use count is spent.
- **Revoked (jti axis)** — lease whose `jti` is in the revocation store.
- **Revoked (key-generation axis)** — lease whose `kgen < keyGenerations[sub]`.
- **Wrong kind** — a non-signing claim (e.g. a secret token) presented where a signing lease is required, or vice versa.

None of these may panic; each is a typed, distinguishable error.

## Acceptance criteria

Each names its test layer.

1. **(unit)** An **expired** lease produces the expired error, no panic.
2. **(unit)** An **exhausted** lease (use count spent) produces the exhausted error, no panic.
3. **(unit/integration)** A **jti-revoked** lease is rejected with the revoked error.
4. **(unit/integration)** A **generation-revoked** lease (`kgen` below the current key generation) is rejected with the revoked error.
5. **(unit)** A **wrong-kind** lease produces the wrong-kind error, no panic.
6. **(integration)** `exp` and `use` are accounted **once**: redeeming and signing decrements the single handle-table counter, and there is no second counter that could diverge — assert by exhausting the use count through the handle and confirming the next attempt is refused.
[...]

PR #330, opened 2026-07-23
Redeem session signing leases through the core handle table

mike-north:
## Summary

Implements the read side of a signing-key lease: `VaultKeeper::redeem_signing_lease`
decrypts a lease JWE (produced by `mint_signing_lease`, #328) and, if it passes every validation
axis, installs an opaque signing capability handle in the core `HandleTable` (#268/#283) — the
same accounting authority `authorize()` already uses for secret tokens. There is no second,
independently-maintained expiry/use counter anywhere on this path.

As part of closing the "wrong kind, or vice versa" axis, `authorize()` now also refuses a
signing-key lease presented as a secret token (`AuthorizationDenied`), rather than silently
storing it as a broken `StoredClaims::Secret` handle whose absent secret made `read_secret` return
a misleading `AccessorConsumed` instead of a typed wrong-kind error.

## Design

- `redeem_signing_lease(host, jwe)` shares its decrypt + key-lookup + rotation-response logic with
  `authorize()` via a new private `decrypt_and_prepare_response` helper — one code path for the
  step every redemption entry point needs, so the two can't drift.
- Validation order: wrong-kind (`kty`) → `validate_claims` (expiry, use-limit, shape) →
  `validate_lease_revocation` (jti + kgen axes) → `HandleTable::record_usage` → install via
  `register_signing_handle`.
- `register_signing_handle` (the existing finite-expiry gate from #282, previously unreachable —
  `#[allow(dead_code)]`) is now redemption's real caller, always with `Some(claims.exp)`.
  `VaultClaims::exp` is a required `u64`, not `Option<u64>`, so there is no decryptable lease
  payload that could route a `None` through to that gate — the never-expiring case stays
  structurally unreachable from this path, not merely policy-refused.
- No new `VaultError` variants: the five axes map onto existing typed variants
  (`TokenExpired`, `UsageLimitExceeded`, `TokenRevoked` ×2, `AuthorizationDenied`), matching the
  repo's existing error taxonomy.

All new tests live in `crates/vaultkeeper-core/src/vault.rs`'s `mint_signing_lease_tests` module
(reuses that module's mock `SigningBackend`/`HostPlatform` test doubles for a full issue→redeem
round trip):

1. (unit) Expired → `redeem_signing_lease_rejects_an_expired_lease`
2. (unit) Exhausted → `redeem_signing_lease_rejects_an_exhausted_lease`
3. (unit/integration) jti-revoked → `redeem_signing_lease_rejects_a_jti_revoked_lease`
4. (unit/integration) generation-revoked → `redeem_signing_lease_rejects_a_generation_revoked_lease`
5. (unit) Wrong kind, both directions → `redeem_signing_lease_rejects_a_secret_token_wrong_kind`,
   `authorize_rejects_a_signing_lease_wrong_kind`
6. (integration) Single accounting authority →
   `redeeming_a_lease_repeatedly_is_accounted_once_by_the_handle_table` (a `use_limit: Some(2)`
   lease redeems twice, both handles stay independently resolvable, the third redemption is
   refused)
7. (unit) Finite-expiry + principal gate stays intact →
   `redeem_signing_lease_cannot_bypass_the_finite_expiry_gate`

Plus a happy-path sanity test, `redeem_signing_lease_returns_a_working_handle_for_a_valid_lease`.

Note on criterion 2/6's expected error: the *first* redemption to hit a lease's `use_limit` also
calls `block_token(jti)` (mirroring `authorize()`'s existing behavior), so the *next* redemption
attempt can observe either `UsageLimitExceeded` or `TokenRevoked` depending on which check
`validate_claims` reaches first — both tests accept either, matching the existing convention in
`authorize_enforces_use_limit` (`crates/vaultkeeper-core/tests/unit_tests.rs`).

## Proof

- `cargo test --workspace`: all green (vaultkeeper-core lib: 409 passed, incl. 10 new tests; full
  workspace: 100% green, no failures)
- `cargo clippy --all-targets -- -D warnings`: clean
- `cargo clippy -p vaultkeeper-wasm --target wasm32-unknown-unknown -- -D warnings`: clean
- `cargo fmt --check`: clean
- `pnpm check` / `pnpm test`: clean (265 passed, 19 pre-existing skips, 0 failed)
- WASM rebuilt with the pinned toolchain (`crates/vaultkeeper-wasm/wasm-toolchain.env` — local
  rustc/wasm-pack/wasm-opt versions matched exactly) and the export/import fingerprint confirmed
  unchanged against the previously-committed artifact (this PR doesn't touch `wasm_impl.rs`'s
  bindings); regenerated `.wasm` committed under `packages/vaultkeeper-wasm/wasm/`. Well within
  the size budget (523 KB uncompressed / 197 KB gzip vs. the 1 MB / 400 KB budgets).

## Deliberate scope decisions

- No CLI (`lease redeem` or similar) or WASM binding wiring in this PR — the issue and its
  acceptance criteria are scoped to the core `VaultKeeper` method and its Rust-level test coverage.
  Wiring a user-facing entry point is a separate, follow-on concern.

Refs #300
Refs #282


mike-north:
Review-thread dispositions (inline replies are blocked while a pending review exists, so recording them here):

- **vault.rs:296 — `\` line-continuation in the error string** (3634968598): Not a bug. Rust's string-literal grammar strips the newline and the following line's leading whitespace after a trailing `\`; we compiled and ran the exact literal and the message renders with no artifacts. Same convention as the pre-existing `wrong_kind_error` in `identity/handles.rs`.
- **vault.rs — `rotated_jwt` built before validation** (3635674119): Fixed in 7d79565. `decrypt_and_prepare_response` is split into `decrypt_claims` + `prepare_response`; both call sites now run full validation before any re-encryption, so error paths never pay the rotation cost.
- **vault.rs:296 — name `authorize()`/`redeem_signing_lease()` in the message** (3635674127): Declining. These messages cross the wasm boundary to TypeScript consumers, where Rust method names would be wrong; the binding-neutral wording is deliberate.
- **vault.rs — `use_limit` not consumed by signing** (3635704122): Real observation, tracked in #345 rather than blocking this PR. No operation in the repo consumes a resolved signing handle to sign yet (`resolve_signing_claims` returns metadata only), and handle repeatability is a documented, pre-existing tradeoff mitigated by the mandatory finite-expiry gate. #345 pins the decision — signing-side budget consumption vs. expiry-only bounding — to land with the first real sign() API.

Also strengthened in this PR per review: AC3/AC4 now discriminate the revocation axis via message substrings (96a243c), since `TokenRevoked` carries no structured axis field.

Issue #333, opened 2026-07-23
Fold exec into run (--token source option, --as var name); deprecate exec until 1.0


mike-north:
## Context

Surface-governance ruling B9 (docs/specs/001, owner-adjudicated): `run` is the single launcher verb — a delegated command gets one or more secrets available in its subshell, and everything else (`--profile`, `--token`, `--set`) only describes how many secrets and the means by which they are populated. Two launcher verbs for one operation is a grammar defect.

## Task

Fold `exec` into `run` and deprecate it:

1. `run` gains `--token <JWE>` as a source option (mutually exclusive with `--profile`/`--profile-file`; combinable with `--set`), injecting the redeemed value under `--as <VAR>` (default `VAULTKEEPER_SECRET` for compatibility with today's `exec`).
2. `run --token` inherits `run`'s full stdio/signal-transparency contract and UATs (child stdout untouched, diagnostics to stderr, signal forwarding, 128+N exit codes) — the transparency contract lives in exactly one command.
3. `exec` becomes a hidden deprecated alias for `run --token`: same behavior, deprecation notice on **stderr** (never stdout), removed at 1.0. Help text for `exec` marked deprecated pointing at `run --token`.
4. Both CLIs while the TS CLI lives; conformance cases updated (exec cases retargeted through the alias path plus new run --token cases).
5. Spec discipline: this PR updates docs/specs/001-surface-governance.md §1.0/§1.1 rows for `run`/`exec` in the same diff (per the E1 evolution rule).

## Dependencies

#279 (run --profile) must merge first — this issue builds on its launcher plumbing. Do not start before then.

## Acceptance criteria

1. `run --token <jwe> -- cmd` matches today's `exec --token <jwe> -- cmd` behavior byte-for-byte on the child's observable environment and stdio (UAT comparing both paths).
2. `--as VAR` renames the injected variable; default `VAULTKEEPER_SECRET`; invalid var names rejected with a typed error.
3. `--token` with `--profile` or `--profile-file` is a usage error naming the conflict.
4. `exec` still works, emits a single-line deprecation notice to stderr only, and its exit codes/stdout are unchanged (UAT asserts stdout byte-identical with and without the notice-bearing stderr).
5. Standard gates green; spec updated in-diff; help snapshots (when they exist) regenerated.

Refs #276.

Issue #279, opened 2026-07-21
Add vaultkeeper run verb with stdio- and signal-transparent child launch


mike-north:
Refs #276

Add the `vaultkeeper run` verb: resolve a profile (via the resolution planner) and launch a child command with the resolved environment, with **full stdio and signal transparency**. Core owns flag semantics, validation, error strings, and `--dry-run` rendering; only argument parsing and the final spawn are per-host.

This is a new verb, not an extension of `exec`. `exec --token <jwe> -- cmd` is token redemption; `run --profile <name> -- cmd` is environment composition (named declaration, minting where needed, launch). Both persist permanently. Do not overload `exec` with an optional `--profile`.

## CLI shape

```
vaultkeeper run --profile <NAME> [OPTIONS] -- <COMMAND>...

  --profile <NAME>       Named profile from $CONFIG_DIR/profiles/<NAME>.json
  --profile-file <PATH>  Load a profile from an explicit path instead
  --set <VAR>=<SECRET>   Ad-hoc rung-2 entry, layered over the profile (repeatable).
                         Marked UNREVIEWED in --dry-run output.
  --dry-run              Print each var, its rung, source backend, and resolved policy, then exit
                         without launching. Never prints values.
  --require-presence     Fail unless every minted entry proved fresh human presence at mint
```

There is deliberately **no** `profile resolve --format env` — a command whose whole job is printing resolved plaintext re-creates the plaintext-in-a-file problem this feature removes.

## stdio and signal transparency (required — this is a correctness requirement, not polish)

MCP stdio servers speak the protocol as framed JSON over stdin/stdout. `run` sits directly in that path.

- **Inherit descriptors; do not pipe and forward.** Native: `std::process::Command` with `Stdio::inherit()` on all three streams. Node: `spawn(cmd, args, { stdio: 'inherit' })`. Inheritance hands the child the parent's real file descriptors, making byte-exactness and zero added buffering true *by construction*. **Never route the child through `HostPlatform::exec`** — that API captures output, and capture breaks MCP stdio.
- **The parent must never read stdin** — any read in the wrapper steals bytes from the child's protocol stream.
- **stdout is the child's alone.** Every vaultkeeper diagnostic — the file-only degradation notice, warnings, progress, errors — goes to **stderr**. Nothing vaultkeeper emits may appear on stdout while a child is running. (`--dry-run` prints to stdout but exits without launching, so there is no conflict.)
[...]

PR #335, opened 2026-07-23
Add vaultkeeper run verb with full stdio and signal transparency

mike-north:
## Summary

Adds a new `vaultkeeper run --profile <NAME> [OPTIONS] -- <COMMAND>...` verb to the native Rust
CLI: resolves a profile (via the #278 resolution planner) plus any `--set` overlay, then launches
the command with full stdio and signal transparency — the wrapper never reads stdin, the child's
stdout is untouched by any vaultkeeper output, and SIGINT/SIGTERM are forwarded rather than
letting the child be orphaned.

`run` is a distinct verb from `exec`, not an extension of it: `exec --token <jwe> -- cmd` is token
*redemption*; `run --profile <name> -- cmd` is environment *composition* (named declaration,
minting where needed, launch).

## Flag surface (exact spelling)

```
vaultkeeper run --profile <NAME> [OPTIONS] -- <COMMAND>...

  --profile <NAME>                Named profile from $CONFIG_DIR/profiles/<NAME>.json
  --profile-file <PATH>           Load a profile from an explicit path instead
  --set <VAR>=<SECRET>            Ad-hoc rung-2 entry, layered over the profile (repeatable)
  --dry-run                       Print the plan, never mint, never launch
  --require-presence-at-issuance  Refuse a real run unless every minted lease entry proved
                                   presence at issuance (not yet enforceable — see below)
```

There is deliberately no `profile resolve --format env`.

## Presence-flag naming rationale (owner-adjudicated, superseded once)

Presence flags follow a scope-suffixed family convention — every presence flag names the
guarantee it demands. `--require-presence-per-use` (existing, `store`/`delete`) forces a fresh
human action *per operation*; `--require-presence-at-issuance` (this `run` verb) demands every
minted lease entry prove presence *at issuance*. These are different guarantees and must not
share a bare `--require-presence` spelling. (The flag was originally named
`--require-presence-at-mint`; "mint" was retired from the vocabulary in favor of "issuance" during
review, and the name was changed throughout — flag definition, help text, dry-run rendering,
tests.)

**Enforcement posture (fail-closed, not a silent no-op or a warn-and-proceed):**
`resolve_profile` (from #278) has no way yet to verify a minted lease entry's `pres` claim at
issuance time — it takes no `HostPlatform` to prompt with, and there is no existing lease being
checked (issuance mints a *new* one). Given that, a real (non-dry-run) `run` invocation with
`--require-presence-at-issuance` set **refuses outright** — typed error, non-zero exit, nothing on
stdout — rather than proceeding silently or with a mere warning, which would let a caller believe
an unproven guarantee held. `--dry-run` still discloses that a real run would refuse (it never
mints or launches regardless, so disclosure there is safe). Real enforcement (checking the `pres`
claim during resolution) is a tracked follow-up, not silently dropped.

## Design notes / deviations

- **Rust CLI only.** `profile`/`session` (the #277/#278 building blocks `run` depends on) exist
  only in the native Rust CLI today — `packages/cli` (the TS CLI) has no profile-resolution
  concept at all yet, only a diagnostic-only WASM export
  (`crates/vaultkeeper-wasm/src/wasm_impl.rs`'s `__testLoadProfile`). Porting profile resolution
  to the TS CLI is out of scope for this issue and would be a large undertaking in its own right;
  `run` follows the same precedent `profile`/`session` already established. Flag semantics,
  `--set` parsing, and `--dry-run` rendering all live in `vaultkeeper-core`
  (`crates/vaultkeeper-core/src/run.rs`) specifically so a TS CLI port can render identical output
  later without re-deriving the rules.
- **No executable-trust binding of the launched command (yet).** `resolve_profile`'s
  `ResolveOptions::executable_hash` binds a verified executable identity into a `materialize:
  "lease"` entry's `exe` claim for `sigstore`/`registry`-tier entries. `run` passes `None` — real
  support would mean resolving `command[0]` through `$PATH` to a real file, hashing it, and
  running it through the same TOFU/Sigstore flow `VaultKeeper::setup()` uses for its caller. Until
  that lands, a `materialize: "lease"` profile entry whose resolved trust tier is
  `sigstore`/`registry` is refused by `resolve_profile` itself (`VaultError::ExecutableTrustRequired`)
  rather than silently minted with an incoherent `"dev"` exe marker — a `materialize: "secret"`
  entry at any trust tier, and a `materialize: "lease"` entry at `minTrust: "unverified"`, both
  work today. Tracked as a natural follow-up, not silently dropped.
- **Source selection is extensible for issue #333.** After this PR lands, `run --token <JWE>
  [--as VAR]` is planned to become a second, mutually-exclusive source alongside `--profile`/
  `--set`, with `exec` deprecating in its favor. `run`'s source selection is shaped as an enum
  (`RunSource`, currently just `Profile { profile, profile_file }`) rather than `--profile`
  hard-wired as the only entry point, and the launcher (`launch_and_wait`/stdio/signal handling)
  only ever depends on a plain `ResolvedEnv` map and a command line — no dependency on `RunSource`
  or profile types at all — so a future `Token` source can slot in and reuse the entire launch
  path unchanged.
- Signal forwarding uses `libc::kill` (added as a small, unix-only dependency of
  `vaultkeeper-cli`) rather than `Child::kill()` (SIGKILL only) or shelling out to `kill(1)`.
- SIGINT/SIGTERM handlers are installed **before** `cmd.spawn()` (not inside the wait loop after
  it) — see `launch_and_wait`'s doc comment — so a signal arriving in the window immediately after
  launch can't hit default disposition and orphan the child.
- The post-`spawn()` `env.zeroize()` scrubs `launch_and_wait`'s own `ResolvedEnv` buffer, not the
  separate copy `tokio::process::Command` already made into its internal env storage for the
  child's lifetime (documented honestly in the source; neither `tokio::process::Command` nor
  `std::ffi::OsString` expose a way to scrub that storage).

All UATs are in `crates/vaultkeeper-cli/tests/run_uat.rs`, run against the real compiled binary
and real child processes (no mocks). Conformance cases are in
`crates/vaultkeeper-conformance/src/lib.rs`'s `run_cases()`.

1. **(UAT) stdio round trip** — `ac1_stdio_round_trip_is_byte_exact_through_a_framed_message_style_payload`:
   pipes a framed-message-shaped payload (including null and non-UTF8 bytes) into `run -- cat`
   and asserts stdout is byte-identical.
2. **(UAT) stdout purity** — `ac2_stdout_contains_exactly_the_childs_bytes_and_the_degradation_notice_is_stderr_only`:
   file-only backend (degradation notice fires) + a child `cat`-ing a file with a known byte
   sequence (embedded newline + a genuine non-UTF8 byte); asserts stdout is exactly those bytes
   and the notice is on stderr only.
3. **(UAT) signals** — `ac3_sigterm_reaches_the_child_and_the_wrapper_outlives_it` and
   `ac3_sigint_reaches_the_child_and_the_wrapper_outlives_it`: send real `kill -TERM`/`-INT` to the
   wrapper's own pid (polling a child-written readiness marker first, closing the fixed-sleep race
   an earlier revision had), assert a trap in the child fired (marker file) and the wrapper's
   `wait()` only returns after the child exits.
4. **(UAT) exit semantics** — `ac4_child_exit_code_is_propagated`,
   `ac4_a_signal_killed_child_yields_128_plus_n` (also readiness-marker-based, not a fixed sleep),
   `ac4_child_closing_its_own_stdout_early_is_not_treated_as_an_error` (renamed from a version that
   overclaimed EPIPE coverage — closing your own write end can't produce EPIPE), plus the new
   `ac4_a_broken_pipe_from_the_downstream_reader_does_not_panic_or_crash_the_wrapper`, which
   genuinely drives EPIPE via the classic `yes | head -1`-style trigger (drop the stdout reader
   while the child streams) and asserts a clean signal-death exit with no panic/backtrace.
5. **(UAT) `--dry-run`** — `ac5_dry_run_prints_plan_and_never_mints_or_launches` and
   `ac5_dry_run_never_mints_a_rung_3_lease_even_when_backend_would_otherwise_refuse`: asserts
   vars/rungs/backend/`UNREVIEWED` markers print, no value ever appears, nothing launches, and a
   lease entry that would be *refused* on a real run (no executable-trust binding yet) still
   renders successfully under `--dry-run` because dry-run never resolves at all. Also
   `run_dry_run_discloses_require_presence_at_issuance_would_refuse_a_real_run` and
   `run_refuses_a_real_launch_when_require_presence_at_issuance_is_set` for the fail-closed
   presence-at-issuance posture.
6. **(conformance)** `run_cases()` in `crates/vaultkeeper-conformance/src/lib.rs`: flag-surface
   validation (`--profile`/`--profile-file` required-and-conflicting, malformed `--set`, `--help`
   documents the exact `--require-presence-at-issuance` spelling), plus — using a new
   `ConformanceCase::extra_files` mechanism (writes real fixture files, e.g. a loadable
   `profiles/<name>.json`, into the isolated config dir before running) — genuine coverage of the
   "No command specified" branch (previously a same-named case only reached profile lookup), a
   real `--dry-run` render, and the `--require-presence-at-issuance` refusal path. Run against the
   real binary from both the Rust test runner and the JS conformance runner
   (`packages/cli-tests/test/conformance/`).

## Test gates (all green locally)

- `cargo test --workspace`: **656 passed, 0 failed**
- `cargo clippy --workspace --all-targets -- -D warnings`: clean
- `cargo fmt --check`: clean
- `pnpm check`: clean (no TS API surface touched by this change)
- `pnpm test`: **1750 passed, 24 skipped, 0 failed** across the workspace, including the JS
  conformance runner picking up all 36 `run` cases from the regenerated `cases.json`
- WASM rebuilt (`wasm-pack build --target nodejs crates/vaultkeeper-wasm`, pinned toolchain per
  `crates/vaultkeeper-wasm/wasm-toolchain.env`); the rebuild's export/import fingerprint matches
  (47/47 exports, 54/54 imports) and the `.wasm` binary itself came out byte-identical to the
  already-committed one this round (no functional wasm-reachable surface changed) — nothing to
  re-commit.
- The `run_uat.rs` signal/exit-code tests were run repeatedly (including under `cargo test`'s
  default parallelism, which surfaced the original fixed-sleep flake) to confirm the
  readiness-marker fix closed it — 5/5 clean runs after the fix, versus reproducible failures
  before it.

Refs #279

## Follow-up

Real `--require-presence-at-issuance` enforcement (verifying every lease-materialized entry's issuance proved presence via the `pres` claim, refusing the specific offending entry) is tracked in #336, which also specs the dedicated typed error variant that replaces the interim `MaterializeModeUnsupported` reuse used here for the refuse-as-unimplemented posture. Refs #336.


mike-north:
Review-thread dispositions (inline replies are blocked while a pending review exists, so recording them here; the fresh pending review at f22cd09 reflects the final state):

**Round-1 review findings (the ten mike-north threads)** — all addressed in 3c31539: signal handlers now installed via `SignalGuard` before `spawn()`; the `Command` env-copy residual is documented honestly instead of overclaimed; dry-run output writes through a non-panicking helper; the shared-foreground-pgrp double-delivery is documented; AC3/AC4 use readiness-marker polling instead of fixed sleeps; a genuine EPIPE test was added and the old one's claim corrected; AC2 now asserts real byte-exactness (embedded newline + non-UTF8 byte); the misnamed conformance case was renamed and real no-command/dry-run cases added; `--profile-file` gained a happy-path test; the help text no longer points at Rust module docs; and `--require-presence-at-issuance` on a real launch now fails closed with a typed error instead of being a silent no-op.

**Copilot threads:**
- Degradation notice keyed on configured backend type (3635096207): real fragility, currently unreachable (file is the only CLI backend today) — tracked in #346.
- `render_dry_run` omits `useLimit` (3635096224): fixed in ebc0d57, mirroring `profile show`'s rendering, with tests.
- Conformance case name vs assertion (3635096239): fixed in 3c31539 (renamed + genuine cases added).
- `extraFiles`/`extra_files` path validation, TS and Rust (3635196905, 3635196925): fixed in ebc0d57 — both runners now reject absolute paths and traversal, via Component inspection on the Rust side.
- Presence-flag spelling vs description (3635196931): stale — the PR description documents the owner-adjudicated rename to `--require-presence-at-issuance`, applied throughout.
- TS validator weaker than Rust (3635690773): fixed in 28f180d — exact Component::Normal parity.
- Undocumented `MaterializeModeUnsupported` slug (3635762564): fixed in cd6c837 — `"run-require-presence-at-issuance"` registered in the errors.rs slug contract.
- Rust validator doc/error wording (3635844459): fixed in 0739d7a — wording now states the full rule, symmetric with TS.
- Env not zeroized on early error returns (3635917689, 3635917718): fixed in f22cd09 — the resolved env is held in `Zeroizing<ResolvedEnvMap>`, so every exit path scrubs; happy-path scrub timing unchanged; regression test added.
- Rust accepts `a//b` where TS rejects (3635917732): fixed in f22cd09 — explicit raw-segment check added alongside Component inspection, with a pre-fix-failing regression test.

The branch was also rebased onto main (a6ecfa3) to resolve the wasm-artifact conflict from #330's merge, with the artifact rebuilt last under the pinned toolchain and verified against the drift guard.

Issue #345, opened 2026-07-23
Wire sign() for session signing leases (redemption-counted Model A accounting)


mike-north:
Refs #300, #261, PR #330.

PR #330 delivers the redemption side of session signing leases: `redeem_signing_lease` validates every axis and installs a finite-expiry handle, with redemption use-accounting done once in the core handle table. What does not exist yet — anywhere in the codebase — is an operation that consumes a resolved signing handle to perform cryptographic signing (`resolve_signing_claims` returns `{kid, backend_ref}` metadata only; the only caller of `create_detached_jws` is a unit test).

Issue #300's AC6 wording ("redeeming and signing decrements the single handle-table counter … assert by exhausting the use count through the handle") anticipates a signing-side counter that has no operation to hang off today. Meanwhile `identity/handles.rs`'s module doc documents the current posture explicitly: signing handles are repeatable, not one-shot, and the mandatory finite-expiry gate (#282) is the enforcement axis.

When the real signing operation lands, decide and implement one of:

1. **Signing consumes the lease budget** — store the lease's `jti`/remaining-use linkage with the handle at redemption and call `record_usage` on each signing use, refusing when spent (keeps the single-accounting-authority invariant; matches AC6's literal wording), or
2. **Only redemption is use-limited** — signing bounded solely by `expires_at` (matches the existing module-doc framing); update #300/AC6-adjacent docs so the contract is explicit.

Either way the decision should land together with (or before) the first `sign()`-style API, so an unbounded-signing window never exists. Also relevant to the invocation-time principal re-check tracked in #261.

Acceptance:
- A written decision (this issue) reflected in `identity/handles.rs` module docs.
- If option 1: handle-table entries carry remaining-use state, signing decrements it, and a test exhausts the budget through actual signing and confirms refusal.
- If option 2: docs updated and a test pins that signing does not decrement redemption budget.

mike-north:
**Decision (2026-07-24): resolved.**

Signing-use accounting for **session** signing leases is **redemption-counted (Model A)**: one redemption spends one unit of `use_limit`, installing a finite-expiry handle that signs non-interactively within the session until it expires. `use_limit` counts *sessions*, not signatures. This is exactly what #330 already built (single decrement at redemption in the core handle table) — no rework, and it keeps signing off the concurrency-hot-path (see #322).

[...]

Issue #261, opened 2026-07-19
Stakeholder question: custody model for an automation agent's signing keypair (delegated signing without human presence)


mike-north:
Filed by the harness program (mike-north/ai-plugins, `docs/harness-program/judge.md`). The program enrolls an automation signer — a reasoning agent ('the judge') with its own Ed25519 keypair, authorized in attest-it for exactly one gate. The key must be usable non-interactively by that agent's harness, but should never be extractable by the agents it governs (a desperate coding agent must not be able to produce the judge's signature — that visibility property is the point of the identity).

**Question**: what is the recommended vaultkeeper custody model for such a key?

- Is delegated signing (per attest-it#122's direction) the intended home — the key held by vaultkeeper, the judge harness granted a signing capability, raw key material never entering the agent's environment?
- Can a delegated-signing grant be scoped so only a specific harness/agent-type context can invoke it (aligning with the 1Password-environment-per-agent-type model in `docs/harness-program/agent-type-harnesses.md`)?
- If the external-key import API (attest-it#122's dependency) isn't near, what's the interim: file-backed key in the judge's 1Password environment?

**Acceptance**: a recommendation comment (or pointer to existing docs) for near-term custody + the target end-state. No urgency until the program's M2 (judge stand-up); filing early so the answer shapes the judge's identity design rather than retrofitting it.

mike-north:
Timing update from the harness program: M2 preparation has begun (the judge line's canon is merged, including the identity-and-enrollment design that explicitly defers custody to this issue's answer — `docs/judge/identity-and-enrollment.md` in mike-north/ai-plugins). No urgency change to the ask itself, but a recommendation in the next few working sessions would let the custody model shape the M2 identity provisioning issue (ai-plugins #109) rather than retrofit it. The interim-fallback design (file-backed key in a scoped 1Password environment) is already written down should delegated signing not be ready in time.

mike-north:
Consumer-side context from the judge line (the agent whose keypair this question is about). The judge's canon design has since landed — `docs/judge/identity-and-enrollment.md` in mike-north/ai-plugins — and it deliberately defers custody to whatever this issue recommends, so the answer shapes the design rather than being retrofitted. Recording the constraints it has to satisfy, in case they narrow the option space:

[...]

PR #347, opened 2026-07-24
Close public HandleTable::insert_signing bypass of the #282 signing-handle gate

mike-north:
## Summary

Issue #282's core ask — gate `VaultKeeper::register_signing_handle` behind an
issuance-side check — was already fully implemented on `main` by #283 (and
reinforced by #330's `redeem_signing_lease`, which routes through the same
gate). That prior work already:

- Rejects `expires_at: None` for signing handles with a typed
  `VaultError::AuthorizationDenied`.
- Narrows `register_signing_handle` to `pub(crate)`.
- Rewords the `crate::identity::handles` module docs and the method doc from
  "long-lived non-interactive signing is fully supported" to "deliberately
  out of scope for now, gated on the issuance-side principal check tracked in
  #261", while explicitly distinguishing non-extractability of key material
  (already handled by `SigningBackend`) from non-invocability by an untrusted
  caller (not yet handled).
- Has a regression test (`register_signing_handle_tests::none_expiry_signing_handle_is_gated`
  and `vault_keeper::redeem_signing_lease_cannot_bypass_the_finite_expiry_gate`)
  proving a `None`-expiry registration is refused.

While verifying the issue's "no public path can mint a never-expiring or
unlimited-use signing handle" acceptance criterion against the *current*
state of `main`, I found one residual gap the prior PRs didn't close:
`HandleTable::insert_signing` was still a plain `pub fn` on `HandleTable`,
which is re-exported from `vaultkeeper-core`'s crate root — and
`vaultkeeper-core` is published to crates.io. That means any downstream
crate could construct a bare `HandleTable::new()` and call
`insert_signing(claims, None)` directly, minting the exact durable ambient
signing capability the `register_signing_handle` gate exists to prevent —
bypassing the gate entirely rather than merely calling it un-gated.

This PR closes that one remaining public path: `HandleTable::insert_signing`
is now `pub(crate)`, consistent with the #283 precedent for
`register_signing_handle` itself. No other crate in the workspace
(`vaultkeeper-cli`, `vaultkeeper-wasm`, `vaultkeeper-conformance`) calls
`insert_signing`, so this is not observably breaking within the repo. I also
fixed the resulting `rustdoc` `private_intra_doc_links` warning on
`HandleId`'s doc comment (it linked to the now-private `insert_signing`).

## Scope note

Per the issue's hard-scope framing, this PR does **not** touch `sign()` or
any signing operation — signing-use accounting is tracked separately in
#345. It also does not implement the issuance-side principal check itself
(#261) — that remains the open product decision this gate exists to wait
for.

1. *"No long-lived / unlimited / never-expiring signing handle class should
   be mintable until an invocation-time principal check exists"* — already
   covered by `register_signing_handle`'s `None`-expiry refusal (#283); this
   PR additionally closes the `HandleTable::insert_signing` bypass so the
   gate can't be sidestepped by a caller that skips `VaultKeeper` entirely.
   Covered by the pre-existing `none_expiry_signing_handle_is_gated` and
   `redeem_signing_lease_cannot_bypass_the_finite_expiry_gate` tests, plus
   the fact that `cargo build`/`cargo test`/`cargo clippy` all pass with
   `insert_signing` now `pub(crate)` and zero external callers.
2. *"Reword the module + method docs..."* — already done on `main` (#283);
   verified unchanged and still accurate by this PR (`handles.rs` module
   docs § "Why `expires_at` must stay caller-supplied... issue #282").
3. *"Distinguish non-extractability of key material from non-invocability by
   an untrusted caller"* — already present in the module docs (#283);
   verified unchanged by this PR.

## Test plan

- [x] `cargo test` (workspace) — all green (425+ core unit tests, 88
      `vault_keeper`/integration tests, plus CLI/wasm/conformance suites).
- [x] `cargo clippy --all-targets --all-features -- -D warnings` — clean.
- [x] `cargo fmt --check` — clean.
- [x] `cargo doc -p vaultkeeper-core --no-deps` — no new warnings (fixed the
      one introduced by narrowing `insert_signing`'s visibility).
- No WASM rebuild needed — `insert_signing`/`HandleTable` are not part of
  the surface `vaultkeeper-wasm` wraps.

Refs #282

### DIFF
--- a/crates/vaultkeeper-core/src/identity/handles.rs
+++ b/crates/vaultkeeper-core/src/identity/handles.rs
@@ -159,7 +159,7 @@ pub const HANDLE_TABLE_MAX_SIZE: usize = 10_000;
 
 /// Opaque handle identifier returned by [`HandleTable::insert_secret`] or by
 /// the crate-internal signing-handle insertion path (see
-/// [`crate::vault::VaultKeeper::register_signing_handle`], the sole gated
+/// `crate::vault::VaultKeeper::register_signing_handle`, the sole gated
 /// entry point for signing handles — issue #282).
 ///
 /// Carries no claims data itself — it is a bare, unguessable (UUID v4)
--- a/crates/vaultkeeper-core/src/identity/handles.rs
+++ b/crates/vaultkeeper-core/src/identity/handles.rs
@@ -157,8 +157,10 @@ use crate::util::time::now_secs;
 /// (release/expiry/usage-limit are).
 pub const HANDLE_TABLE_MAX_SIZE: usize = 10_000;
 
-/// Opaque handle identifier returned by [`HandleTable::insert_secret`] /
-/// [`HandleTable::insert_signing`].
+/// Opaque handle identifier returned by [`HandleTable::insert_secret`] or by
+/// the crate-internal signing-handle insertion path (see
+/// [`crate::vault::VaultKeeper::register_signing_handle`], the sole gated
+/// entry point for signing handles — issue #282).
 ///
 /// Carries no claims data itself — it is a bare, unguessable (UUID v4)
 /// lookup key. **It is bearer capability material, not a public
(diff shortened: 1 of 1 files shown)


mike-north:
Review complete. One issue found and fixed in-branch: the PR claimed to fix the rustdoc `private_intra_doc_links` warning, but the `HandleId` doc comment had merely retargeted its link to `VaultKeeper::register_signing_handle` — which is also `pub(crate)` — so `cargo doc -p vaultkeeper-core` still warned. Fixed in 438c329 by using a plain code span; verified the warning is gone and `cargo test -p vaultkeeper-core` (88 tests), `clippy -D warnings`, and `fmt --check` are green.

Verified independently: `insert_signing` has zero callers outside the crate; no other public path on `HandleTable` can mint a `Signing` entry (only `insert_secret`/`insert_signing` reach the private `insert`); the `register_signing_handle` None-expiry gate and its regression tests are untouched. A compile-fail test for the visibility narrowing would be low-value (the compiler enforces it) and was deliberately not added.

Note: one pre-existing instance of the same warning class remains at `handles.rs:106`, predating this PR — left alone as out of scope.

Issue #313, opened 2026-07-22
Build the StubTool paired-double framework and extend the conformance corpus with world-mutation, state-conditional, and shape-assertion axes


mike-north:
Refs #288
Refs #306

Build the shared corpus + fake-CLI framework that is the spine of the automated-test-maximization strategy, and extend the existing conformance corpus so **one corpus runs against two targets**. The suite-parameterized-over-target shape decided here is load-bearing for every item after it.

## The organizing commitment: one contract corpus, two targets, per backend

Every production backend adapter has a **paired double**, and its manual residue collapses to one mechanically-answered question: "does the real adapter still align with its double?" The mechanism is **one contract suite, two targets** — a shared per-backend corpus runs against (a) the **double** in CI, every PR, headless, and (b) the **real adapter** occasionally, with the real account/hardware present. **Alignment *is* "both targets green on the same corpus."** When the real tool changes behavior, the real-target run fails and the diff names precisely what the double must learn — drift is a corpus divergence, never a discovery, never a matter of judgment. The suite-over-target is *the deliverable*; the doubles are its instruments. This item builds the double-side instrument for the subprocess-driven backends and the shared corpus both targets run.

### 1. A generic `StubTool` fake-CLI harness (subprocess-driven backends)

One generic fake-CLI harness, configured per tool — **not N hand-written stubs**. Every subprocess-driven target (`security`, `op`, `ykman`, `secret-tool`, `powershell`) presents the same observable contract: **argv + stdin in → stdout + stderr + exit code out, against a shape-faithful mutable world** (real item objects, **not** a flattened `id → secret` map). A single stub binary reads a **behavior table** (match on subcommand + args + stdin → emit stdout/stderr/exit, mutate the typed world) and covers all five; per-tool differences are **data, not code**. (The 1Password DesktopAuth dylib path is a different mechanism — an in-process flavored double — tracked separately.)

- Lives in a small **Rust crate `vaultkeeper-stub-tools`** that compiles to the fake binaries, plus a thin JS shim — both consumers need it. Rust core tests spawn the fake binary through the real backend's `HostPlatform::exec`; the JS conformance runner (`packages/cli-tests`) puts the same binary on a test-scoped `PATH`. One binary, two harnesses, identical tables — this keeps Rust-side and JS-side coverage honest against each other.
- **Seed:** #293's stub-ykman already has a match/emit shape; generalize it into `StubTool` and port stub-ykman onto the generalized framework as the proof case.

[...]

Issue #311, opened 2026-07-22
Add PresenceSimulatorBackend to @vaultkeeper/test-helpers with stacked production-unreachability guards


mike-north:
Refs #307

Add a distinct `PresenceSimulatorBackend` to `@vaultkeeper/test-helpers` that lets tests control vaultkeeper's presence signal — most critically simulating its **absence**, so a consumer can prove in CI that an automation signer attempting a presence-gated operation is **refused**. The class is structurally isolated from production by three stacked guards.

## Motivation and the hard constraint

The consuming project is designing a gate predicate along the lines of "this gate requires at least one presence-backed signature": certain gates must not be satisfiable by an automation signer (an agent keypair) alone, because an agent cannot tap a YubiKey. The security property that matters is the **negative** one — "an automation signer attempting to satisfy a presence gate is refused" — and it is untestable without producing, on demand in CI, a signing attempt that lacks presence.

**A presence simulator is a presence forger.** If it can be reached from a production code path, it *is* the vulnerability that presence-backed gates exist to prevent. A convenience flag on the normal API is the worst possible outcome — it would hand every consumer a presence bypass in exchange for test convenience. The stakeholder's stated position: they would rather have no test helper than one that can be switched on in production. This isolation requirement governs the shared delivery surface.

### `PresenceSimulatorBackend`

- Implements the existing `PresenceCapableBackend` vocabulary from `packages/vaultkeeper/src/backend/types.ts` — **no parallel concept**; align capabilities and per-operation outcomes with the existing `BackendCapabilities` vocabulary.
- Capabilities and per-operation outcomes are **scriptable: grant / refuse / not-capable.**
- The **negative case falls out of vaultkeeper's existing fail-closed enforcement** rather than being separately simulated: a signer whose backend does not advertise presence for the operation is refused by the vault itself with a typed `NotCapableError`, **before the backend is touched** (the `enforce_presence_requirement` fail-closed pattern). The simulator only makes both sides of that boundary drivable in CI.

### Three stacked unreachability guards

1. **Structural** — lives only in `@vaultkeeper/test-helpers` (a devDependency, never in any runtime dependency graph) and is **never registered in the backend registry** (`register-builtins.ts`), so no `config.json` on any machine can name it and no production vault instance can load it. The only wiring path is direct construction in test code.
[...]

Issue #293, opened 2026-07-21
Port YubikeyBackend to vaultkeeper-core (versioned GCM, legacy-CBC detection, ykman orchestration)


mike-north:
Refs #288

Port the YubiKey backend (`YubikeyBackend`) into `vaultkeeper-core` as a subprocess-orchestration + filesystem + crypto backend driving `ykman` over `HostPlatform::exec`. This is the largest of the four ports: it combines subprocess orchestration, an on-disk encrypted store with its own metadata, and a versioned GCM crypto format with legacy detection.

## argv safety (already clean)

Per the argv audit, `ykman` invocations are already argv-clean: **no secret ever reaches the `ykman` argv** — only id-derived hex challenge values do. The secret is encrypted at rest under a key derived from the YubiKey's HMAC-SHA1 challenge-response, and only the derived/hex material crosses the process boundary. Preserve this; the argv-sentinel regression test still applies (it must confirm no secret in any child argv, alongside the other three subprocess backends).

## Crypto + on-disk format

- Add the `hkdf` crate (HKDF key derivation off the challenge-response material). Justify against the repo's lean-dependency posture in the PR; this same `hkdf` crate is shared with the core `SigningBackend` work.
- Reimplement the versioned `1:iv:tag:ct` GCM envelope, including **legacy-CBC detection** for blobs written by older versions — decrypt legacy-CBC blobs on read, re-seal as versioned GCM.
- Preserve the `metadata.json` layout (id list / per-entry metadata) exactly as the TS backend writes it.
- YubiKey's storage directory requires `mkdir(0o700)`; use the `HostPlatform` directory-creation path rather than a `.keep` sentinel workaround.

## On-disk format compatibility

Blobs and `metadata.json` already written by the TS backend must remain readable by the core port. Validate against TS-written fixtures, following the existing precedent under `crates/vaultkeeper-core/tests/fixtures/ts-written-keystate/`.

## Acceptance criteria

1. (unit) An argv-sentinel regression test confirms the secret appears in **no** child `ykman` process argv (only id-derived hex values do).
2. (unit) The versioned `1:iv:tag:ct` GCM envelope round-trips; a **legacy-CBC** blob (captured as a fixture) is detected, decrypted, and re-sealed as versioned GCM on read.
3. (unit) `metadata.json` written by the core port matches the TS backend's layout; a `metadata.json` + blob set written by the TS backend is read correctly by the port (TS-fixture compat).
[...]

PR #348, opened 2026-07-24
Build the StubTool paired-double framework and extend the conformance corpus

mike-north:
## Summary

Builds the shared corpus + fake-CLI framework described in #313: one contract corpus, driven against two targets (an in-process/subprocess double now, the real adapter occasionally per the release hardware matrix), for subprocess-driven backends.

- **`vaultkeeper-stub-tools`** (new crate): a generic `StubTool` engine — a `BehaviorTable` (pure data: argv/stdin match rules → emitted stdout/stderr/exit + world mutation) driven against a shape-faithful `World` (real `Item { id, fields: BTreeMap<String, String> }` objects, not a flattened `id -> secret` map). Five `vk-stub-*` binaries (`vk-stub-security`, `vk-stub-op`, `vk-stub-ykman`, `vk-stub-secret-tool`, `vk-stub-powershell`) are thin two-line wrappers around the same shared `cli::run_stub_process` — no per-tool Rust branches anywhere in the engine.
- Ports issue #293's `stub-ykman` (a hand-written `match`/`emit` `TestHost::exec` in `crates/vaultkeeper-core/src/backend/yubikey.rs`'s tests) onto the generalized framework — same 29 tests, now delegating to `tables::ykman()` + `StubTool::run`.
- Extends `crates/vaultkeeper-conformance`'s `ConformanceCase` with a new, backward-compatible `stub_scenario: Option<StubScenario>` field carrying the three new axes (world-mutation, state-conditional-response, shape-assertion) as an ordered `Vec<StubStep>` run against one shared `World`. Every pre-#313 case is untouched (`#[serde(default)]`, existing runners keep working unmodified); the new axis cases live in their own `stub_scenario_cases()` function, consumed only by the two new StubTool-axis test files.
- Guardrails: a stub refuses to run without a test-only env sentinel (`VAULTKEEPER_STUB_TOOLS_SENTINEL`); binaries are named `vk-stub-<tool>` and only ever resolve as the real tool name via a symlink inside a test-scoped `PATH` dir; no interactive-prompt code path exists anywhere in the engine (grant/deny/timeout always resolves from the table); `PresenceSimulatorBackend` (#311) remains untouched and is not referenced here.

1. **(unit, Rust)** `StubTool` reproduces the contract for ≥2 distinct tool shapes, table-driven, no per-tool branches — `crates/vaultkeeper-stub-tools/tests/engine_table_driven.rs` (`ykman_*`, `secret_tool_*`).
2. **(integration, Rust)** Fake binary spawned through a real backend's `HostPlatform::exec` — `crates/vaultkeeper-cli/src/host.rs`'s `stub_tool_secret_tool_ac2` module: `SecretToolBackend` + `NativeHostPlatform` against the real, compiled `vk-stub-secret-tool` binary resolved via a test-scoped `PATH` symlink.
3. **(integration, JS)** Same fake binary resolved from a test-scoped `PATH` by `packages/cli-tests`, byte-identical to the Rust harness for the same table — `packages/cli-tests/test/conformance/stub-tool-axes.test.ts`, paired with the Rust-side `crates/vaultkeeper-stub-tools/tests/conformance_axes.rs` driving the identical `stub_scenario_cases()` corpus in-process.
4. **(unit)** `ConformanceCase` gains all three axes; a store→get round trip, a locked-vs-unlocked conditional, and a field-structure round-trip assertion all parse and drive the stub, with existing conformance assertions unchanged — `vaultkeeper_conformance::stub_scenario_cases()` + `crates/vaultkeeper-stub-tools/tests/conformance_axes.rs`.
5. **(integration)** Shape-assertion case FAILS on a shape-mismatched write (reproducing the op-spike silent-field-drop defect) and PASSES on a shape-faithful write — `crates/vaultkeeper-stub-tools/tests/shape_assertion.rs` (`shape_mismatched_stdin_fails_the_round_trip_assertion_reproducing_the_op_spike_defect` is `#[should_panic]`, `shape_faithful_stdin_passes_the_round_trip_assertion` passes, both call the same `assert_shape_round_trips` helper).
6. **(integration)** #293's stub-ykman ported onto `StubTool`, existing cases pass — `crates/vaultkeeper-core/src/backend/yubikey.rs`'s `TestHost::exec` now delegates to `tables::ykman()` + `StubTool::run`; all 29 pre-existing yubikey tests pass unchanged.

## Scope notes

- `security` and `powershell` get minimal probe-only tables (`vk-stub-security`/`vk-stub-powershell` exist, are table-driven, and compile/run) rather than full round-trip coverage — `KeychainBackend` drives `security -i` with a multi-command stdin script and `DpapiBackend` drives PowerShell, both meaningfully more complex shapes than the five ACs above require to prove out. `secret-tool` and `ykman` (real, wired-in Rust backends) and `op` (the shape-assertion demonstration table, matching the issue's motivating defect) carry the actual round-trip/shape coverage.
- The `op` table models the real `op item get --format json` field-array shape directly (no live `op` backend exists in this repo yet) purely to reproduce the motivating defect class by construction.
- CI: added a drift check (`export-stub-scenarios` vs. the committed `stub-scenarios.json`, mirroring the existing `export-backend-conformance` pattern) and a `vk-stub-tools-binaries` artifact upload/download so the TS job can run the JS StubTool-axis suite against the real binaries without a Cargo toolchain.
- No production (non-test/non-dev-dependency) code changed in `vaultkeeper-core` or `vaultkeeper-wasm` — `vaultkeeper-stub-tools` is only ever a dev-dependency of `vaultkeeper-core`/`vaultkeeper-cli`, and the only touched `vaultkeeper-core` code is inside `#[cfg(test)]`. No WASM rebuild needed.

## Proof

- `cargo test` (workspace): green.
- `cargo clippy --all-targets -- -D warnings` (workspace): clean.
- `cargo fmt --check`: clean.
- `pnpm check` (typecheck + lint + api-report + format, all packages): green.
- `pnpm test` (full TS suite including `packages/cli-tests`, `stub-tool-axes.test.ts`, `run-conformance.test.ts`): green.

Refs #313


mike-north:
Review complete across four lenses (Rust, TypeScript, tests, cross-cutting). The framework held up well:

- The engine is fail-loud where it matters: an unmatched invocation exits 127 with a descriptive stderr (never silent success), matching is first-rule-wins and documented, all state is BTreeMap-ordered (no iteration-order leakage), and the shape-assertion test discriminates its panic message so it can't pass on an unrelated setup failure.
- The 'no production code changed' claim was verified precisely: stub-tools is dev-dependency-only, publish = false, and every touched line in core/cli is inside #[cfg(test)]. No packaging path can sweep the stub binaries into a release artifact.
- The ported ykman stub preserves the old TestHost contract (29 tests unchanged), and the stub-scenarios drift check plus same-run artifact scoping in CI are sound.

Fixed in-branch (through a588b3a, including the rebase over main):
- The JS runner's output matcher was a stripped copy that silently dropped JsonContains and the Rust (?s)-to-dotall translation — the exact divergence the parity harness exists to prevent, latent until the first scenario used either. Both suites now share one matcher. [open thread on stub-tool-axes.test.ts — addressed]
- The sentinel-absent and unknown-table refusal paths (the security guardrails) had zero coverage — two subprocess tests added against the real binary.
- world.rs linked a nonexistent Mutation variant — corrected, cargo doc is warning-free; the op() table doc now describes the real unresolved-placeholder mechanism rather than claiming an empty field; CLAUDE.md's crate list and workspace tree now include the new crate.
- Rebased over main and carried the run --token conformance cases across the stub_scenario schema extension (regenerated, not hand-merged).

Noted without change: AC3's 'byte-identical' phrasing is stronger than what's directly asserted (both harnesses pass the same corpus rather than diffing each other's raw bytes) — the in-code comment already states the weaker, accurate contract; a direct cross-harness byte-diff would be reasonable future hardening. The Rest-must-be-last table invariant is debug_assert-only, acceptable while tables are compile-time literals.

Issue #326, opened 2026-07-22
YubiKey delete() claims presence coverage it never enforces (TS + Rust lockstep)


mike-north:
## Context

Found in #324's review (inherited from the TS backend, not introduced by the port): `YubikeyBackend.get_capabilities()` reports `presence_enforced_operations: None` (= every keyed operation forces a touch) when `require_touch` is set, but `delete()` never invokes challenge-response — only a device presence probe plus a filesystem delete — so no touch is ever demanded for deletion. A caller trusting `presence_per_use` to mean "deleting this secret requires a physical touch" is misled. Identical behavior in `packages/vaultkeeper/src/backend/yubikey-backend.ts`.

## What

Decide the honest contract in lockstep across both implementations: either (a) perform a challenge-response touch before delete (real enforcement), or (b) report `presence_enforced_operations: Some([Read, Store, Sign?])` excluding Delete, letting `enforce_presence_requirement` fail closed for `--require-presence-per-use` deletes per #258's operation-scoped design. Option (b) matches the #258 machinery's purpose. Update capability docs and parity tests both sides.

Refs #324, #258.

mike-north:
Picking this up: making the YubiKey delete() presence contract honest in TS/Rust lockstep via option (b) from the issue — report operation-scoped `presence_enforced_operations` excluding Delete so `--require-presence-per-use` deletes fail closed per #258's design, rather than bolting a challenge-response onto delete(). Capability docs and parity tests updated on both sides, with regression tests that fail against today's behavior. PR will reference this issue with Refs.

mike-north:
PR opened: https://github.com/mike-north/vaultkeeper/pull/351

PR #351, opened 2026-07-28
Fix YubiKey delete() overclaiming presence-per-use coverage

mike-north:
## Summary

`YubikeyBackend.get_capabilities()`/`getCapabilities()` reported `presence_enforced_operations: None` (Rust) / omitted `presenceEnforcedOperations` (TS) whenever the configured slot enforces touch — which the shared capability contract reads as "every keyed operation is covered." But `delete()` never performs the HMAC-SHA1 challenge-response that actually demands the physical tap; it only probes device presence (`require_device()`/`requireDevice()`) and unlinks the entry file. A caller relying on `presencePerUse` to mean "deleting this secret requires a physical touch" was misled, and a `--require-presence-per-use` delete silently succeeded with no touch.

This implements option (b) from the issue: both implementations now report `presenceEnforcedOperations` scoped to `read`/`store` only (the two operations that actually invoke challenge-response), excluding `delete`. `sign` does not apply — this backend does not implement signing. This lets `enforce_presence_requirement`/`#enforcePresenceRequirement` fail closed for a `--require-presence-per-use` delete via the operation-scoped machinery from #258, instead of silently succeeding.

1. **Rust `get_capabilities()` excludes Delete when `require_touch` is set** — `crates/vaultkeeper-core/src/backend/yubikey.rs::tests::get_capabilities_excludes_delete_from_presence_enforced_operations_issue_326`
2. **TS/Rust lockstep contract** — same shape (`['read', 'store']` / `[Read, Store]`, omitting delete) implemented in both `crates/vaultkeeper-core/src/backend/yubikey.rs` and `packages/vaultkeeper/src/backend/yubikey-backend.ts`; covered by `packages/vaultkeeper/test/unit/backend/capabilities.test.ts::YubiKey excludes delete from presenceEnforcedOperations (regression, issue #326)` mirroring the Rust test above
3. **`--require-presence-per-use` delete fails closed** — Rust: `crates/vaultkeeper-core/src/backend/yubikey.rs::tests::require_presence_per_use_delete_fails_closed_issue_326`. TS: `packages/vaultkeeper/test/integration/presence-enforcement.test.ts::YubikeyBackend delete is not presence-enforced (regression, issue #326) > a --require-presence-per-use delete against a touch-configured YubiKey backend fails closed with NotCapableError` — both drive the real `YubikeyBackend`/`get_capabilities` through the shared enforcement path (`enforce_presence_requirement`/`VaultKeeper.delete`), not a hand-built approximation. Enforcement runs before any backend touch, so no real device/`ykman` is required.
4. **Regression tests referencing #326, failing pre-fix** — the four named tests above all assert `Some([Read, Store])` / `['read', 'store']` (Rust/TS) and a `NotCapable`/`NotCapableError` refusal; against unmodified code `presence_enforced_operations` was `None`/omitted, so `enforces(Delete)` was `true` and the enforcement test would have observed the delete proceed rather than being refused.
5. **Cross-implementation parity** — the Rust unit-test suite in `crates/vaultkeeper-core/src/backend/yubikey.rs` and the TS suites in `packages/vaultkeeper/test/unit/backend/capabilities.test.ts` / `packages/vaultkeeper/test/integration/presence-enforcement.test.ts` are the existing mirrored (ported) test suites for this contract (see the module-level doc comments referencing each other) and were extended in lockstep with equivalent regression tests in both languages.
6. **Capability docs updated** — doc comments on both `PresenceCapableBackend for YubikeyBackend` impls now explain the operation-scoped rationale; `docs/manual-tests/presence-per-use.md` YubiKey section gained a new manual-verification step (5) documenting that `delete` is not touch-enforced even with a required touch policy.
7. **Standard gates green** — `pnpm check`, `pnpm test`, `cargo test --workspace`, `cargo clippy --workspace --all-targets`, `cargo fmt --all -- --check`, and `pnpm exec prettier --check` on touched files all pass locally.

## Notes

- No public API surface change: `YubikeyBackend` is `@internal` in the TS package, so `pnpm generate:api-report`/`generate:api-docs` produced no diff (`pnpm check`'s `check:api-report` passed as part of `pnpm check`).
- Per the dispatch brief, changes are confined to the yubikey backend files, their tests, and capability docs — no edits to `types.rs`, `errors.ts`, or CI workflow files.

Refs #326, #258, #324

### DIFF
--- a/crates/vaultkeeper-core/src/backend/yubikey.rs
+++ b/crates/vaultkeeper-core/src/backend/yubikey.rs
@@ -50,7 +50,7 @@ use zeroize::{Zeroize, Zeroizing};
 use crate::backend::file::hex_encode;
 use crate::backend::types::{
     BackendCapabilities, ExecOptions, HostPlatform, ListableBackend, PresenceCapableBackend,
-    SecretBackend,
+    PresenceOperation, SecretBackend,
 };
 use crate::errors::VaultError;
 
@@ -589,13 +589,31 @@ impl ListableBackend for YubikeyBackend {
 /// `getCapabilities` exactly: the answer always comes from the
 /// operator-declared configuration, never derived from the backend type
 /// alone.
+///
+/// **Operation coverage (issue #326):** the touch-per-operation policy only
+/// actually fires for `store()` and `retrieve()` — both perform the
+/// HMAC-SHA1 challenge-response (`ykman otp calculate 2 ...`) that demands
+/// the physical tap. [`SecretBackend::delete`] only calls `require_device()`
+/// (a presence *probe*: is a YubiKey plugged in at all) plus a filesystem
+/// unlink — it never invokes challenge-response, so it never actually
+/// demands a fresh touch. Reporting `presence_enforced_operations: None`
+/// (meaning "all keyed operations") would therefore be dishonest: a caller
+/// relying on `presencePerUse` to mean "deleting this secret requires a
+/// physical touch" would be misled. This reports
+/// `Some([PresenceOperation::Read, PresenceOperation::Store])`, excluding
+/// `Delete`, so a `--require-presence-per-use` delete fails closed via
+/// [`crate::vault::enforce_presence_requirement`] instead of silently
+/// succeeding with no touch. `Sign` does not apply — this backend does not
+/// implement `SigningBackend`.
 #[cfg_attr(target_arch = "wasm32", async_trait::async_trait(?Send))]
 #[cfg_attr(not(target_arch = "wasm32"), async_trait::async_trait)]
 impl PresenceCapableBackend for YubikeyBackend {
     async fn get_capabilities(&self) -> Result<BackendCapabilities, VaultError> {
         Ok(BackendCapabilities {
             presence_per_use: self.require_touch,
-            presence_enforced_operations: None,
+            presence_enforced_operations: self
+                .require_touch
+                .then(|| vec![PresenceOperation::Read, PresenceOperation::Store]),
         })
     }
 }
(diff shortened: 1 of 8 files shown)


mike-north:
Review complete across four lenses (Rust, TypeScript, tests, cross-cutting). The fix itself verified sound end-to-end:

- Confirmed empirically that `store`/`retrieve` invoke the tap-demanding challenge-response and `delete` does not, in both implementations — `['read', 'store']` is the factually correct enforced set, and `enforce_presence_requirement`/`#enforcePresenceRequirement` genuinely fail closed for a flagged delete with no TOCTOU between capability check and operation.
- All four regression tests verified to fail against pre-fix code and to drive the real enforcement path (Rust: `enforce_presence_requirement` directly; TS: `VaultKeeper.delete` end-to-end). 31 Rust yubikey tests and 29 TS tests pass.
- Sibling backends checked for the same overclaim pattern: FileBackend/InMemoryBackend report `presencePerUse: false`; 1Password already lists its exact set. Nothing else newly misleading.

One defect found and fixed in-branch (deb747d): the shared `BackendCapabilities.presenceEnforcedOperations` JSDoc still cited a touch-configured YubiKey as the canonical "omitted = everything enforced" example — which this PR just falsified. Reworded and regenerated `docs/api/`.

One suggestion, comment-only: the new TS test `YubiKey excludes delete from presenceEnforcedOperations (regression, issue #326)` fully overlaps the adjacent exact-shape `toEqual` assertion updated in the same file (the Rust suite avoided this by keeping its pre-existing test untouched and adding one new one). Harmless duplication; left as-is to preserve the AC-to-test-name mapping in the PR description.

